PR #161, opened 2015-01-05
Add setup instructions for Amazon EC2 instances

pfeyz:


gureckis:
awesome... i'm merging in without testing myself, but assume that people in need of this will point out if there are any issues.  Thanks... I suggest for now that we start urging people towards this solution instead of OpenShift since RedHat did something to break the psiturk standalone process.

gureckis:
Thanks @pfeyz !!

hstojic:
Thanks a lot for the instructions! 

I've just tried it out and works smoothly. The only thing was that I couldn't find the instance at first (ubuntu-psiturk-2 - ami-bcab37d4), but when I switched to US East (N.Virginia) there it was. I googled a bit and it seems that only an owner can copy the instance to other locations. Perhaps that's unnecessary but then we should update the instructions.

RoyalTS:
This works great for the psiturk example task but I'm running into a few issues with the custom task that I want to run and that runs smoothly on my Macbook:

- the version of pip on the image seems to be busted and cannot be used to install any packages. The accepted answer in this SO thread seems to help: http://stackoverflow.com/questions/27341064/how-do-i-fix-importerror-cannot-import-name-incompleteread
- I would love to run the dev version of psiturk (mostly because I need access to Flask sessions and the version in master doesn't set a SECRET_KEY) but am not nearly unix ninja enough to pull this off. Can I simply git clone everything that's in the dev branch into /usr/local/lib/python2.7/dist-packages/psiturk? Or is there a better way to do this?

RoyalTS:
On startup and in response to several commands psiturk also throws the following warning:

    : InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
      InsecurePlatformWarning

This can be fixed by running

    sudo pip install requests[security]

The better fix, apparently is upgrading to a more current version of Python 2.7 (>=2.7.9) but that's not an option in this case because apparently every version of Ubuntu is fine-tuned to a particular version of Python.

pfeyz:
It looks like the version of requests that's pre-installed from the ubuntu  on the image works, but if you install requests via pip you end up installing the too-new version.

    ubuntu@ip-xxx:~/debug2$ pip --version
    pip 1.5.4 from /usr/lib/python2.7/dist-packages (python 2.7)

    ubuntu@ip-xxx:~/debug2$ sudo pip install --upgrade requests
    Downloading/unpacking requests from https://pypi.python.org/packages/py2.py3/r/requests/requests-2.6.0-py2.py3-none-any.whl#md5=3ab1972bbaf2802d94516fb86b9b0d0b
      Downloading requests-2.6.0-py2.py3-none-any.whl (469kB): 469kB downloaded
    Installing collected packages: requests
      Found existing installation: requests 2.2.1
        Uninstalling requests:
          Successfully uninstalled requests
    Successfully installed requests
    Cleaning up...

    ubuntu@ip-xxx:~/debug2$ pip --version
    Traceback (most recent call last):
    ....
    ImportError: cannot import name IncompleteRead

Of course now you can't uninstall it using pip, but you can remove it by force:

    $ python -c "import requests; print requests.__file__"
    /usr/local/lib/python2.7/dist-packages/requests/__init__.pyc

    $ sudo rm -r /usr/local/lib/python2.7/dist-packages/requests

    $ python -c "import requests; print requests.__file__"
    /usr/lib/python2.7/dist-packages/requests/__init__.pyc

    $ pip --version
    pip 1.5.4 from /usr/lib/python2.7/dist-packages (python 2.7)

And now we're importing the apt-get-installed version of requests ($ sudo apt-get install python-requests)

pfeyz:
As for installing the dev branch, pip can install [from a git branch](http://psiturk.readthedocs.org/en/latest/install.html?#install-directly-from-github).

    > Can I simply git clone everything that's in the dev branch into 
    > /usr/local/lib/python2.7/dist-packages/psiturk?

You shouldn't need to attempt this given what I mentioned, but no, this wouldn't work. You'd want to clone the dev branch and then run

    $ sudo python setup.py install

from within it. Using [virtual environments](http://psiturk.readthedocs.org/en/latest/install.html?highlight=virtualenv#running-inside-a-virtual-environment) makes it easier to switch between psiturk versions.

pfeyz:
I'm unable to reproduce the "InsecurePlatformWarning" issue.

RoyalTS:
@pfeyz Thanks a bunch! I've got everything running smoothly now and suspect the InsecurPlatformWarning issue was the result of my nixing the pre-installed version of pip as per the linked SO thread. Your solution is much cleaner.

PR #170, opened 2015-02-19
remove superfluous code

RoyalTS:
Best as I can tell (I'm still gettting my bearings in psiturk's codebase so apologies if I'm wrong) this doesn't actually do anything. sys.path.append(os.getcwd()) doesn't throw ImportErrors and at least in your whoops y'all example the contents of custom_models.py are imported in custom.py, not in experiment.py

### DIFF
--- a/psiturk/experiment.py
+++ b/psiturk/experiment.py
@@ -75,12 +75,6 @@ except ImportError:
 else:
     app.register_blueprint(custom_code)
 
-try:
-    sys.path.append(os.getcwd())
-
-except ImportError:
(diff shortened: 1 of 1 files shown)


gureckis:
yeah I think this is a correct assessment...

gureckis:
thanks for these patches!  btw, for future reference, we prefer patches to go to the dev branch rather than master, but I just added these anyway.

RoyalTS:
Got it! The other projects I work on always patch master and then branch for releases so I just lazily assumed. Will send patches against dev in the future.

PR #172, opened 2015-03-10
cache control max age measured in seconds, not ms

RoyalTS:
see http://flask.pocoo.org/docs/0.10/config/

Also: Any reason the cache timeout shouldn't also be user configurable (perhaps with a sensible default)?


gureckis:
no, one could definitely add this as a config option... is there a reason it useful to change from the current value?

PR #181, opened 2015-04-30
setup.py adds sample .psiturkconfig to user's home directory

ehenry2172:
I think it is useful to have a sample .psiturkconfig placed into the user's home directory during installation. I'm a new user and this pull request comes from my installation experience. I didn't follow the quickstart guide, so the long version (http://psiturk.readthedocs.org/en/latest/amt_setup.html) has you manually creating/editing the file rather than an example being autogenerated. Based on this experience, I believe adding the default config file at setup is advantageous for two reasons: 1) It assures the user psiTurk is installed properly (I SEE a necessary file in my home directory - it must have worked!) and 2) It helps new users that didn't follow the quickstart more intuitively edit the configuration file so they can see the syntax. I simply copied the example_global_config_defaults.txt and had it placed into home directory during install as .psiturkconfig. Hope this is helpful.

Thanks!

### DIFF
--- a/psiturk/default_configs/.psiturkconfig
+++ b/psiturk/default_configs/.psiturkconfig
@@ -1,0 +1,8 @@
+[AWS Access]
+aws_access_key_id = YourAccessKeyId
+aws_secret_access_key = YourSecretAccessKey
+aws_region = us-east-1
+
+[psiTurk Access]
+psiturk_access_key_id = YourAccessKeyId
+psiturk_secret_access_id = YourSecretAccessKey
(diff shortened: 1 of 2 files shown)


gureckis:
my concern is that it could be viewed as bad to install things in the home directory on install.  i would much rather this happen on *first launch* of psiturk (does it not already?).  an example is that ssh exists on most unix systems but the ~/.ssh folder isn't created until you actually use the ssh command.  i think that is more typical for the unix world.

PR #187, opened 2015-06-25
ignore other users processes when checking server

alexanderrich:
Currently, when checking the server psiTurk searches through ALL user's
processes for psiturk server processes, and thus if a second user is
running a psiturk server it will find these processes and try to
communicate with them, causing an error. If psiturk only searches for
the current user's processes, multiple users should be able to
serve from the same server, as long as they use different ports.


alexanderrich:
@jbmartin  here's a pull request! simply removes the flag on the "ps" command that causes other user's processes to be listed as well. This has been informally tested with me and anselm serving simultaneously from burn.

PR #194, opened 2015-08-12
update total HIT cost based on new fees

amandayung:
Since Amazon changed its pricing structure on July 22nd, it may be good to report the actual cost of HITs when you create them, rather than reporting the cost with the old 10% commission fee. The change made takes into consideration the number of assignments (20% for < 10 assignments, 40% for >= 10).


gureckis:
:+1:

Issue #142, opened 2014-08-29
Error message at end of experiment


sweetticket:
I'm always getting this error message every time a participant finishes the experiment. However, I think all of my data is still being saved on MySQL (I checked via phpMyAdmin). I'm using OpenShift with MySQL.

Traceback (most recent call last):
  File "/var/lib/openshift/53d9c3855973ca578a0003ee/python/virtenv/lib/python2.7/site-packages/flask/app.py", line 1817, in wsgi_app
    response = self.full_dispatch_request()
  File "/var/lib/openshift/53d9c3855973ca578a0003ee/python/virtenv/lib/python2.7/site-packages/flask/app.py", line 1477, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/var/lib/openshift/53d9c3855973ca578a0003ee/python/virtenv/lib/python2.7/site-packages/flask/app.py", line 1381, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/var/lib/openshift/53d9c3855973ca578a0003ee/python/virtenv/lib/python2.7/site-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/var/lib/openshift/53d9c3855973ca578a0003ee/python/virtenv/lib/python2.7/site-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/var/lib/openshift/53d9c3855973ca578a0003ee/app-root/runtime/repo/psiturk/user_utils.py", line 11, in new_func
    resp = make_response(f(_args, *_kwargs))
  File "/var/lib/openshift/53d9c3855973ca578a0003ee/app-root/runtime/repo/psiturk/experiment.py", line 321, in start_exp
    filter(Participant.workerid == workerId).\
  File "/var/lib/openshift/53d9c3855973ca578a0003ee/python/virtenv/lib/python2.7/site-packages/sqlalchemy/orm/query.py", line 2241, in all
    return list(self)
  File "/var/lib/openshift/53d9c3855973ca578a0003ee/python/virtenv/lib/python2.7/site-packages/sqlalchemy/orm/query.py", line 2353, in **iter**
    return self._execute_and_instances(context)
  File "/var/lib/openshift/53d9c3855973ca578a0003ee/python/virtenv/lib/python2.7/site-packages/sqlalchemy/orm/query.py", line 2368, in _execute_and_instances
[...]

Issue #197, opened 2015-09-14
multiple processes on same server by different users 


annacoenen:
Currently "is_server_running" in experiment_server_controller.py checks for all current psiturk processes (including by other users and other ports), and so only one person at a time can run psiturk off the same server. Can we somehow filter processes uniquely by users in that function?

Wildcarde:
Ideally, a single user should be able to run multiple experiments as well as long as they are on different ports.

deargle:
I think this is possible now.

PR #198, opened 2015-09-23
Support multiple servers and some other minor bugfixes

rbtrace:
This patchset will allow multiple servers (same user or different users) to run on one host.  This is issue #197.

It'll also fix issue #142.

There are some other minor enhancements in here as well to handle some error conditions and slightly bad behavior.

### DIFF
--- a/psiturk/db.py
+++ b/psiturk/db.py
@@ -10,7 +10,9 @@ config.load_config()
 
 DATABASE = config.get('Database Parameters', 'database_url')
 
-engine = create_engine(DATABASE, echo=False) 
+config_pool_recycle = config.getint('Database Parameters', 'pool_recycle')
+
+engine = create_engine(DATABASE, echo=False, pool_recycle=config_pool_recycle)
 db_session = scoped_session(sessionmaker(autocommit=False,
                                          autoflush=False,
                                          bind=engine))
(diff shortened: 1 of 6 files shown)


alexanderrich:
Thanks for working on these fixes, they look great. One strange bug I'm getting: whenever I try to call the cmdline() method on a psutil Process object (as is done in commit 59f82c3) I get an AccessDenied error. Stranger still, if you call the as_dict() method instead, it works and the dict for each process actually includes 'cmdline' as one of the fields, so access to this information doesn't *really* seem to be restricted. This happens trying to launch the server in psiturk, and also just playing around with psutil, versions 3.2.2 an 2.2.1. I've tried it on two computers, both running OSX.

Have you ever run into this bug? Any idea what's going on?

rbtrace:
My dev environment is Linux so I didn't see this behavior with cmdline()  I just got an OSX environment setup and it appears as though calling cmdline() on kernel processes will raise the AccessDenied exception.  My list comprehension calls cmdline() on all of the processes so this'll get the kernel processes too and raise the exception.

psutil's as_dict() method for the object catches these exceptions and replaces them with None (or anything you want if if you call it like as_dict(as_value='somethinghere').  So that's a cute way to get around the exception issue.

A slight rework of my cmdline processing logic should fix the issue.  Something along the lines of:

...
if PROCNAME in ' '.join(x.as_dict(ad_value='')['cmdline'])

... should do it.

I'll try to test it out tomorrow.

alexanderrich:
@rbtrace ever test out this fix? I was just starting to work on the multiple server issue myself and saw/remembered this old pull request.

PR #207, opened 2016-02-18
get_workers gets more than 100 workers/assignments

### LINKED ISSUE #203: worker list returning limited list, affecting bonuses
I recently ran a HIT in which I was bonusing participants en masse. The n of the study is 119 now but when I run:

worker list --hit [HITID] 

it only returns the first 100 workers. I tested this with some old 100+ participated HITs we had and noticed the same problem. Only the first 100 workers are returned.

Further in our recent study when I try to bonus the 101st and above participants I am unable to. I currently have 19 participants I will need to bonus manually. I was using worker bonus --amount [amount] --hit [HITID] to do this but since it only reads the first 100 participants it is unable to bonus them.

Looking at the worker list source code it looks like the list is pulled amazon's amt_services. Is there currently an issue in accessing worker lists from amazon?

deargle:
get_workers now counts the total number of results for
a given hit, and then requests multiple pages of up to 100
assignemnts per page until all results are obtained.

Note, however, that the code is currently still limited to
returning workers from the first 20 hits.

Fixes #203

### DIFF
--- a/psiturk/amt_services.py
+++ b/psiturk/amt_services.py
@@ -396,13 +396,33 @@ class MTurkServices(object):
             hits = self.mtc.search_hits(sort_direction='Descending',
                                         page_size=20)
             hit_ids = [hit.HITId for hit in hits]
-            workers_nested = [
-                self.mtc.get_assignments(
+           
+            workers_nested = []
+            page_size=100
+            for hit_id in hit_ids:
+                current_page_number=1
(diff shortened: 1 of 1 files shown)


deargle:
Todo for myself, make this code less tacky. Review the cool way [boto](https://github.com/boto/boto/blob/b5be4791f1c74173e663d6ebaef024a341f536c9/boto/mturk/connection.py#L311-L326) does get-all's

PR #209, opened 2016-02-18
get_workers get all hits

### LINKED ISSUE #208: get_workers limited to only returning 20 hits
[the search_hits call in get_workers](https://github.com/NYUCCL/psiTurk/blob/663c2fdb0535f79ff0aec142ead4e0e594c7dc79/psiturk/amt_services.py#L396-L397) is limited to only returning 20 results.

This can be fixed by calling boto's [get_all_hits](https://github.com/boto/boto/blob/054305da314d34f5a7ef8518b3d79af84bf2aea4/boto/mturk/connection.py#L311) and exhausting the iterator

deargle:
get_workers now called boto's get_all_hits(). One consequence of this edit
is that hits returned are no longer sorted descending by creation date.
this will change the order that the returned workers appear in

Fixes #208

### DIFF
--- a/psiturk/amt_services.py
+++ b/psiturk/amt_services.py
@@ -393,8 +393,7 @@ class MTurkServices(object):
         if not self.connect_to_turk():
             return False
         try:
-            hits = self.mtc.search_hits(sort_direction='Descending',
-                                        page_size=20)
+            hits = self.mtc.get_all_hits()
             hit_ids = [hit.HITId for hit in hits]
            
             workers_nested = []


deargle:
Bump, here's why this is important. The code which is supposed to return all hits currently calls `search_hits`, but this only returns the first 20 hits. I am planning on running a study that has _many_ hits for one study, to take advantage of the "< 9 assignments per hit == lower commission rate". So I'm going to have a lot of hits that I might want to see all of the workers for.

This PR switches the code to call boto's `get_all_hits` instead of `search_hits`. Boto docs for get_all_hits says: "Despite what search_hits says, it does not return all hits, but instead returns a page of hits. This method will pull the hits from the server 100 at a time, but will yield the results iteratively, so subsequent requests are made on demand."

gureckis:
thanks!

PR #210, opened 2016-02-18
for bonusing, only selects an assignment_id if completion status >= 4

### LINKED ISSUE #190: Bonus Fails
It occasionally happens that a worker finishes a psiturk task, but fails to "complete" the HIT. Then another worker will take up the assignment and complete the HIT. The result is that there will be two records in the database with the same assignment ID, but with different worker IDs. In this case, attempts to bonus the second worker may fail, although they were successfully approved.

It looks like amt_services.bonus_worker pulls the first record with that assignment id, and fetches the associated worker_id:

```
assignment = self.mtc.get_assignment(assignment_id)[0]
worker_id = assignment.WorkerId
self.mtc.grant_bonus(worker_id, assignment_id, bonus, reason)
```

If the first record fetched is the wrong one, it would, presumably, raise an Exception since the worker_id and assignment_id are not paired with respect to awt. Changing the assignment ID, or deleting one of the records resolves the issue, confirming diagnosis.

deargle:
Fixes #190

### DIFF
--- a/psiturk/psiturk_shell.py
+++ b/psiturk/psiturk_shell.py
@@ -737,7 +737,7 @@ class PsiturkNetworkShell(PsiturkShell):
                 init_db()
                 part = Participant.query.\
                        filter(Participant.assignmentid == assignment_id).\
-                       filter(Participant.endhit != None).\
+                       filter(Participant.status >= 4).\
                        one()
                 if auto:
                     amount = part.bonus


deargle:
Oh goodness I just realized that this includes commits from another branch. I'll redo this.

deargle:
I rebased, but note that more is needed to actually close this issue. See the discussion in #190. This commit doesn't hurt though. Anyone feel free to add to this PR.

deargle:
- [ ] My code needs to be changed to exclude status == 6.

deargle:
Too old and code changed dramatically in refactor, probably.

PR #213, opened 2016-02-29
Add descriptions for each of the errors

### LINKED ISSUE #89: more helpful error messages
when something goes wrong for a worker, they get a messages like "Sorry, error #1005".   Neither the experimenter nor the worker can make sense of this.  One proposal is to document the error codes in the docs website and provide direct links from the error pages to the docs so easy to figure out what it all means.  Currently, when there's an error I have to login to github find experiment_errors.py and look it up.

jhamrick:
Fixes #89 and adds a bunch of descriptions to go with each of the errors. There were a few errors that didn't seem to be used anywhere, so I didn't really know what description to put with them:

* intermediate_save
* ad_not_found
* template_unsafe
* in_debug
* unknown_error

I gave them empty strings for now but really there are I think two options which would be better:

1. Remove these errors since they are no longer used
2. If they should be kept in, give them a non-empty string for the description

### DIFF
--- a/psiturk/example/templates/error.html
+++ b/psiturk/example/templates/error.html
@@ -39,6 +39,7 @@
                         </p>
                         <p>
                             Error: #{{ errornum }}<br />
+                            Description: {{ errordesc }}<br />
                             {% if hitId %} HITID: #{{ hitId }}<br /> {% endif %}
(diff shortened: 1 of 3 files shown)


gureckis:
woo hoo!

PR #214, opened 2016-02-29
Allow repeat participants

### LINKED ISSUE #90: option to allow repeats
currently psiturk defaults to database-level blocking of workers such that a worker who appears in the database can't take the task again.  in some cases this is not desired behavior.  we should create a config option 'allow_repeat' which toggles this.  also, might be good to allow option 'always_show_instructions' because sometimes if a worker has already done a task once they don't need to read the instructions part again.

jhamrick:
Fixes #90 though it doesn't do anything about toggling whether the instructions are shown or not. I figured that should probably be a separate PR. I also added tests to make sure that the toggle works as expected and either blocks or does not block participants from completing the experiment.

Note that this will still block participants who started the experiment but quit early, regardless of the setting of `allow_repeats`.

### DIFF
--- a/doc/config/hit_configuration.rst
+++ b/doc/config/hit_configuration.rst
@@ -19,6 +19,7 @@ like this:
 	psiturk_keywords = stroop
 	organization_name = New Great University
 	browser_exclude_rule = MSIE, mobile, tablet
+    allow_repeats = false
 
 
 `title` [string]
@@ -162,3 +163,13 @@ standard computers (sort of the opposite to the `mobile` and `tablet` exclusions
 Finally `bot` tries to exclude web spiders and non-browser agents like
 the Unix curl command.
 
+`allow_repeats` [boolean]
+-------------------------
+`allow_repeats` specifies whether participants may complete the experiment more
+than once. If it is set to `false` (the default), then participants will be
+blocked from completing the experiment more than once. If it is set to `true`,
+then participants will be able to complete the experiment any number of times.
+
+Note that this option does not affect the behavior when a participant starts
+the experiment but the quits or refreshes the page. In those cases, they will
+still be locked out, regardless of the setting of `allow_repeats`.
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -83,6 +83,7 @@ project:
 	psiturk_keywords = stroop
 	organization_name = New Great University
 	browser_exclude_rule = MSIE, mobile, tablet
+    allow_repeats = false
 
 	[Database Parameters]
 	database_url = sqlite:///participants.db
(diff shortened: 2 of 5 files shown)


gureckis:
OMG, thanks @jhamrick.  This has been on my todo list forever.  However, I think it might be a little more complex for some.  The ad is actually served up over the ad server for most people (maybe not the way you use psiturk?).  In that case I think I have to  make a similar change on the ad.psiturk.org server to support this.  I'm happy to move away from the centralized ad server but only when it becomes more straightforward and documented how to deploy a psiturk process on a cloud-based SSL signed machine.  More or less the issue is that the ad server runs a function similar to the advertisement() function you altered.  This hits /check_worker_status on your local process and tries to see if the subject has run the task before before.  It currently isn't aware of the fact the we might want to let people repeat and thus will block them at the level of the ad if they appear in the database table.   I think my todo list suggested that there was a need to tell the ad server you want to allow repeats and add changes to that function.

Unfortunately, I also haven't made much progress on open-sourcing the ad server itself so people can see this part either so I'm the main bottleneck here for sure.  Will see if I can meet you half way and get the repeat option implemented on the ad server.

jhamrick:
Ah, I see. I do use psiTurk with the ad server but I didn't think to test this PR against an ad created there.

>  I'm happy to move away from the centralized ad server but only when it becomes more straightforward and documented how to deploy a psiturk process on a cloud-based SSL signed machine. 

One possibility here is to move away from the centralized service by either supporting deploying psiTurk on external on-demand hosting services (e.g. Heroku) or just having people deploy on their own servers. We might be able to facilitate this as well by using a service such as [Let's Encrypt](https://letsencrypt.org/about/) which I think will issue free SSL certificates automatically -- in which case we wouldn't even need to have people manually sign up for one.

yuchuochuo1023:
Hi,

I wonder if the field 'allow_repeats' works now? How could I set this field? I didn't find this option in local or global configuration file...

deargle:
@yuchuochuo1023 are you comfortable with git? psiturk does now have the functionality to bypass the ad server (see [here](https://psiturk.readthedocs.io/en/latest/config/shell_parameters.html#use-psiturk-ad-server-true-false) and [here](https://psiturk.readthedocs.io/en/latest/openshift.html)), so you could try this out as long as you weren't using the ad server. I just created [a new branch](https://github.com/NYUCCL/psiTurk/tree/jhamrick-allow-repeats) that resolves the conflicts and merges this PR. You're welcome to install from that branch and bang on things.

deargle:
I merged this into master, but warning to all, it's not yet implemented into the psiTurk ad server.

PR #218, opened 2016-03-22
fixes "MySQL Server has gone away" error for long-term experiments

kachergis:
MySQL defaults to closing connections that have been inactive for 8
hours, which will result in the “MySQL Server has gone away” error if
you have a long-term, low-traffic experiment (e.g., if you want to
leave a server up for a one-session-per-day experiment). Setting the
pool_recycle=3600 flag in the create_engine call forces SQLAlchemy to
hourly refresh DBAPI connections.


jhamrick:
:+1: this is great! I had noticed this issue before and wondered what the cause was.

PR #220, opened 2016-04-11
Adds browser_exclude_rule for Safari

### LINKED ISSUE #219: Trying to exclude Safari also excludes Google Chrome
If you look at `userAgent` for a Chrome browser it will also contain the word "Safari". Here's a screenshot example thanks to http://whatsmyuseragent.com/.

screen shot 2016-04-08 at 7 35 57 pm

As far as I can tell, this is because Google Chrome also uses Apple Web Kit.

This is a problem given the current psiTurk interface for excluding browsers. Since the exclusion is just done based on pattern matching, it's not possible to exclude Safari without also excluding Google Chrome.

The recommendation on [Stack Overflow](http://stackoverflow.com/q/12799212/2571049) is just to check for the both the word "Chrome" and the word "Safari" if you want to target Chrome, but this doesn't really work with the psiTurk interface through specifying something in the `browser_exclude_rule` inside `config.txt` if one wants to exclude Safari but not Google Chrome.

adamliter:
This fixes #219. As disscussed in #219, Google Chrome userAgent strings
include the word Safari. Because of this, trying to exclude Safari also
excludes Google Chrome.

The value for Safari userAgent strings just includes Safari and not
Google Chrome. Therefore, this problem can be fixed by first checking if
both "Chrome" and "Safari" are in the userAgent string. If this is true,
then we do not want to exclude it, so we do nothing. If this check
returns False, then we still want to check to see if "Safari" is in the
userAgent string because that is what we want to exclude. If it is, then
we exclude it.

### DIFF
--- a/psiturk/experiment.py
+++ b/psiturk/experiment.py
@@ -219,6 +219,11 @@ def advertisement():
                (myrule == "pc" and user_agent_obj.is_pc) or\
                (myrule == "bot" and user_agent_obj.is_bot):
                 browser_ok = False
+        elif (myrule == "Safari" or myrule == "safari"):
+            if "Chrome" in user_agent_string and "Safari" in user_agent_string:
+                pass
+            elif "Safari" in user_agent_string:
+                browser_ok = False
         elif myrule in user_agent_string:
             browser_ok = False
 


psiturk/experiment.py
```diff
@@ -219,6 +219,11 @@ def advertisement():
                (myrule == "pc" and user_agent_obj.is_pc) or\
                (myrule == "bot" and user_agent_obj.is_bot):
                 browser_ok = False
+        elif (myrule == "Safari" or myrule == "safari"):
+            if "Chrome" in user_agent_string:
+                browser_ok = True
```
jhamrick: I don't think you want to set `browser_ok` to True, here, because if someone has the rule "chrome,safari", then `browser_ok` will first be set to `False` (for Chrome) and the `True` (because it detects it as Safari, but then Chrome).

adamliter:
@jhamrick Good catch. Thanks, I didn't think of that! :smile: I've updated the pull request, which should address the case that you pointed out. If you could take a look and let me know, I'd appreciate it!

jhamrick:
Yep I think that looks good. 👍  from me @gureckis

gureckis:
I was about to add this onto the ad server to test...  but do we have to worry about `"Safari"` and `"safari"`?  the initial part of the `elif` cares about this case difference but not the later sections.

deargle:
It's always `"Safari"` with a capital S in the Safari user agent string: http://www.useragentstring.com/pages/Safari/

BTW [this history of the user agent string](http://webaim.org/blog/user-agent-string-history/) is an amazing read.

adamliter:
@gureckis The spelling of "Safari" that shows up in the userAgent string is capitalized. However, an end-user of psiTurk might not know this. That's why the conditional checks for both "Safari" and "safari".

For what it's worth, the check of `elif myrule in user_agent_string` is case sensitive, so, for example, if someone enters "chrome" in the `browser_exclude_rule` list, it will still be possible to use Google Chrome. This is because the spelling that shows up in the userAgent string is "Chrome" and not "chrome".

gureckis:
ok, i added this code snippet to the sandbox ad server...  anyone have an easy test to verify it is ok?  we might want to update the docs about the case issue.

deargle:
You could test it using this [user agent string changer](https://chrome.google.com/webstore/detail/user-agent-switcher-for-c/djflhoibgkdhkhhcedjiklpkjnoahfmg?hl=en-US)

An excerpt from that history article:

> And Apple built Safari, and used KHTML, but added many features, and forked the project, and called it WebKit, but wanted pages written for KHTML, and so Safari called itself Mozilla/5.0 (Macintosh; U; PPC Mac OS X; de-de) AppleWebKit/85.7 (KHTML, like Gecko) Safari/85.5, and it got worse.

>And Microsoft feared Firefox greatly, and Internet Explorer returned, and called itself Mozilla/4.0 (compatible; MSIE 8.0; Windows NT 6.0) and it rendered good code, but only if webmasters commanded it to do so.

>And then Google built Chrome, and Chrome used Webkit, and it was like Safari, and wanted pages built for Safari, and so pretended to be Safari. And thus Chrome used WebKit, and pretended to be Safari, and WebKit pretended to be KHTML, and KHTML pretended to be Gecko, and all browsers pretended to be Mozilla, and Chrome called itself Mozilla/5.0 (Windows; U; Windows NT 5.1; en-US) AppleWebKit/525.13 (KHTML, like Gecko) Chrome/0.2.149.27 Safari/525.13, and the user agent string was a complete mess, and near useless, and everyone pretended to be everyone else, and confusion abounded.

adamliter:
@gureckis Are HITs deployed to the AMT sandbox routed through the psiTurk ad server? I could deploy a quick test HIT to the sandbox and see if it works if that's the case.

gureckis:
yep, the AMT sandbox is a great test for it

deargle:
Yes, they are routed through.

adamliter:
Works for me:

screen shot 2016-04-26 at 12 43 56 pm

You can go [here](https://workersandbox.mturk.com/mturk/searchbar?selectedSearchType=hitgroups&searchWords=Language+Learning+Experiment+%28note%3A+experiment+takes+at+most+20+minutes%29) if you want to try it on your own machines.

gureckis:
Ok, I added it to both the sandbox and regular ad server.  Closing this for now, although seeing the error message if could be nice edit the error message a little.  The person might be blocked for using Firefox but the ad text suggests they should download it.  How about:

```
**Unsupported Browser**
Hi there! It looks like you're using a browser which does not properly display 
this task or a device that is not allowed (e.g., mobile phone, tablet, or incompatible 
browser).  You can reach out to the experimenter via the email below to see 
what you can do in order to participate in this HIT.  Be sure to know you exact 
browser type when emailing the requester.
```

deargle:
Awesome I noticed the same thing. That's a good solution for now. A
user-set error message would be awesome but it'd be a bit of work.

gureckis:
ok i updated the error message text on the two servers

adamliter:
👍😀

Issue #180, opened 2015-04-28
Contribute features from Wallace back into psiTurk


suchow:
In the process of creating Wallace, an experimental platform for simulating cultural evolution online, we've made a number of improvements and extensions to psiTurk. I've opened this issue to start a conversation about what you'd be interested in us contributing back to psiTurk proper.

Here's what we've done:
- [ ] Make it possible to run psiTurk on Heroku
- [x] Improve install speed by installing gnureadline only on OS X
- [x] Allow longer HTTP requests
- [x] Create a route that receives  REST notifications from Amazon about HIT-related events
- [x] Create "HIT types" when creating new HITs.
- [x] Allow HIT durations that are < 1 hr.

@jhamrick, @thomasmorgan, @michaelpacer, @smeylan.

jhamrick:
I think the second is already done in #158.

twiecki:
Heroku definitely would make deployment easier.

gureckis:
This all sounds pretty awesome.  I think any/all of these could hit be good pull requests.  One question thought: why is Heroku easier than say OpenShift or AWS?

suchow:
Okay, that pull request has everything except for Heroku support, which I'll do separately — I may need to work on it a bit to make it automatically detect if you're on Heroku and run that code conditionally.

mshvartsman:
+1 for sub-1hr HITs. I think this makes it clearer to worker that the task we give them shouldn't take very long (and therefore is worth doing for the relatively small amount of $ we pay). 

I implemented one way of doing this in https://github.com/mshvartsman/psiTurk/tree/180-sub-hour-tasks. Integers passed in still get read as hours, but `hours:minutes` gets parsed. 

Should I submit a pull request? Should I do it against dev as per the contrib guidelines or master as per #193?

andrewheusser:
@suchow: was heroku support added or did that not make it in?

suchow:
@andrewheusser 

I'm going to close this now because everything on the original list is finished except for Heroku support. A first draft of that support is implemented in https://github.com/suchow/psiTurk-heroku, though I'm not sure how those changes interact with other services that people use to run psiTurk.

Issue #165, opened 2015-02-05
hit creation example has invalid syntax


grbritz:
In the documentation on hit create, it shows setting time limit of 1:30, but when trying to create a new hit with that time, the error "*** durations must be a whole number" is shown.
![screen shot 2015-02-05 at 10 31 24 am](https://cloud.githubusercontent.com/assets/1458268/6066481/282fe24c-ad22-11e4-8303-5dbaefa97ef2.png)

gureckis:
Sorry, we have a slight problem where the docs website pulls from the dev branch (so new docs release quickly) but most people might be using master.  In dev this syntax works, in master, durations must be a whole number (e.g., "1"() and you can't have fractions of hours.

Issue #137, opened 2014-08-20
HIT Duration < 1 hr?


JSlote:
It would be great to have a way to specify hit duration in hours:minutes rather than just hours. I'll get on this and hopefully submit a pull request today or tomorrow.

adamliter:
Is there any possibility of getting this feature on the stable branch, or does the development branch really need to be installed in order to use it?

adamliter:
By the way, could somebody explain to me what the difference between `cutoff_time` in `config.txt` and the duration that you set when you create a hit is? It seems to me like they are the same thing, but I'm guessing I might be misunderstanding something.

If they are the same thing, it seems odd that you can set the same thing in two different ways and that you can set them to be inconsistent.

So I assume they're not the same thing? What's the difference between them? Thanks! :smile:

adamliter:
For anyone wondering the same thing that I was immediately above, this has been addressed in [the Google group thread called "cutoff vs HIT duration"](https://groups.google.com/forum/#!topic/psiturk/v24ndasQy5U):

> anything in the [Sever Parameters] refer to the gunicorn server.  So the cutoff_time there refers to (I believe) how long a socket will stay open waiting without data being passed.  it is not related to anything on AMT

deargle:
@adamliter I'm still a github noob but I think this thread is invisible since it's closed unless username mentions are included. @gureckis the docs advertise that this feature works but it wasn't carried over into master code. If this is the right way to do it, I'm going to try a pull request that re-incorporates @JSlote 's commits into the current master branch.

deargle:
I should have poked around a bit more and thought a bit more before commenting. #180 made it so that fractional hours work, using a decimal format. I'll just PR the docs to point that out.

Also see #165

adamliter:
@deargle With the stable release of psiTurk on PyPI (2.1.2), the decimal format does not work:

```
https://psiturk.org
 ______   ______     __     ______   __  __     ______     __  __
/\  == \ /\  ___\   /\ \   /\__  _\ /\ \/\ \   /\  == \   /\ \/ /
\ \  _-/ \ \___  \  \ \ \  \/_/\ \/ \ \ \_\ \  \ \  __<   \ \  _"-.
 \ \_\    \/\_____\  \ \_\    \ \_\  \ \_____\  \ \_\ \_\  \ \_\ \_\
  \/_/     \/_____/   \/_/     \/_/   \/_____/   \/_/ /_/   \/_/\/_/

[...]

PR #224, opened 2016-04-21
Describe fractional duration functionality

deargle:
Reflects fractional duration functionality introduced in #180. See also #137, #165.


jhamrick:
👍

Issue #121, opened 2014-07-29
improve/allow https experiment workflow


pfeyz:
I needed to run psiTurk over https since our IRB dictates we'd use encryption.
- There are no options in the config to set up Gunicorn with ssl certificates for https.
- It seems like the ad server always directs users to the ip and port specified by the config `host` and `port` vars, or port 80 if it's running on OpenShift, always prefixed by http.Thus the ad server always sends workers to the http version of the experiment. 

While OpenShift supports https automatically on your application domain, (https://python-yourUsername.rhcloud.com, my understanding is that that endpoint is a proxy server that redirects traffic to your particular "gear") I was unsuccessful at getting OpenShift to redirect all http traffic to https https://www.openshift.com/kb/kb-e1044-how-to-redirect-traffic-to-https.

My solution was to (on EC2)
- hard-code port 80 where psiTurk tells the ad server what port it's running on https://github.com/NYUCCL/psiTurk/blob/master/psiturk/psiturk_shell.py#L942 
- run psiTurk listening on port 22362
- run nginx listening on port 80 (http), where it would redirect to port 443 (https)
- run nginx listening on port 443 and acting as a proxy to psiTurk on 22362. http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_pass

This works, but it required forking and hard-coding variables in psiTurk. Allowing users to specify http/https (ssl = on or something) is possible solution, and would work particularly well with OpenShift since they'd get https automatically. I'm not sure where the http prefix is being added right now. Perhaps on the api server?

pfeyz:
Gunicorn has since added [SSL config settings](http://docs.gunicorn.org/en/stable/settings.html#ssl). 

Using [Let's Encrypt](https://letsencrypt.org/) to automate SSL cert generation/signing could make the entire SSL setup process a single psiTurk server command, only requiring that the user is running their server at a domain name.

I'm taking a look at their [ACME](https://letsencrypt.github.io/acme-spec/) [client](https://github.com/letsencrypt/letsencrypt/blob/master/letsencrypt/client.py) to see what this integration would look like.

deargle:
[...]

PR #226, opened 2016-04-22
Use own ad server

deargle:
Okay here goes. This is functionality for toggling the use of the psiturk ad server. Look in the new comments in psiturk/default_configs/local_config_defaults.txt for an overview of the changes. See [this gist][the gist] for an example config. Here's a summary of the process:

1. You can say whether you want to use psiturk's ad server. Default is `true`.
2. If you don't, then just say so and also set an `ad_location` url in the config for the location of your psiturk ad route. This url should point to a proxy server running in front psiturk/gunicorn. I used nginx. See [the gist referenced above][the gist] for a config that works for psiturk (and for flask/gunicorn/nginx in general). Your proxy server will need to handle SSL. You will need to get your own certs right now if you want to do this.
3. You can optionally also have gunicorn work with SSL. If you want to to do this, set paths to both `keyfile` and `certfile` in the `[Server Parameters]` section. Assuming your process has permissions necessary to read the keyfile, they'll be loaded into gunicorn (see [the same gist as before][the gist] for an example of launching psiturk with sudo while still using the proper virtualenv binaries). If you do this, make sure that you switch the url of your proxy_pass in nginx to use the `https://` protocol.

[the gist]: https://gist.github.com/deargle/5d8c01660a77b8090a2cd24efcda2c59

Some other changes:

1. I also copied the slightly-updated styles.css from psiturk.org to the static/css files so that formatting on the "thanks-mturksubmit.html" file matches what happens if using the psiturk ad server. Sorry for the spacing creating lots of line changes, blame vim auto-formatting.
2. I'm bringing back `closepopup.html` from the dead. Also bringing back the old `thanks.html` and calling it `thanks-mturksubmit.html` so that it has the nice submit button thing. Leaving the 2.0 `thanks.html` alone.

Even if this isn't merged, it demonstrates how to go about the process.

Also see #121

### DIFF
--- a/psiturk/example/static/css/style.css
+++ b/psiturk/example/static/css/style.css
@@ -11,80 +11,67 @@ body {
 
 /* text styling */
 strong {
-    font-weight: 600;         
+  font-weight: 600;         
 }
 
 .warm {
-    color: steelblue;
-    text-align: justify;
+  color: steelblue;
+  text-align: justify;
 }
 
 .cool {
-    color: #ccccff;
+  color: #ccccff;
 }
 
 h1 {
-    margin-top: 0px;
+  margin-top: 0px;
 }
+
 
 /* these are all the same for now, but easy to modify individually */
 #container-ad, #container-error, #container-exp, #container-consent, #container-instructions, #container-questionnaire {
-    position: absolute;
-    top: 0px; /* Header Height */
-    bottom: 0px; /* Footer Height */
-    left: 0px;
-    right: 0px;
-    padding: 100px;
-    padding-top: 5%;
-    border: 18px solid #f3f3f3;
     background: white;
-}
-
-#container-consent, #container-exp {  /* override padding to fit long consent forms and exp display*/
-    padding: 10px;
-    padding-left: 20px;
-    padding-right: 20px;
-    padding-top: 1%;
+    margin: 30px;
 }
 
 /* ad.html  - the ad that people view first */
 #adlogo {
-    float: right;
     width: 140px;
     padding: 2px;
     border: 1px solid #ccc;
+    margin-right: 1em;
 }
 
 /* consent.html - the consent form */
 #consent {
-    margin: 0 auto;
-    margin-top: 0;
-    width: 800px;
-    font-size: 100%;
+  margin: 0 auto;
+  margin-top: 0;
+  width: 800px;
+  font-size: 100%;
 }
 
 .legal {
-    text-align: justify;
-    font-size: 11pt;
-    line-height: 1.1em;
+  text-align: justify;
+  font-size: 11pt;
+  line-height: 1.1em;
 }
 
 #trial {
-    margin: 0 auto;
-    width: 800px;
-    text-align: center;
+  margin: 0 auto;
+  width: 800px;
+  text-align: center;
 }
 
 #query {
-    color: darkgray;
+  color: darkgray;
 }
 
 .question {
-    margin-bottom: 30px;
+  margin-bottom: 30px;
 }
 
 /* errors */
 #error img {
-    padding-right: 10px;
-    padding-top: 15px;
+  padding-right: 10px;
+  padding-top: 15px;
 }
--- a/psiturk/example/static/css/style.css
+++ b/psiturk/example/static/css/style.css
@@ -11,67 +11,80 @@ body {
 
 /* text styling */
 strong {
-  font-weight: 600;         
+    font-weight: 600;         
 }
 
 .warm {
-  color: steelblue;
-  text-align: justify;
+    color: steelblue;
+    text-align: justify;
 }
 
 .cool {
-  color: #ccccff;
+    color: #ccccff;
 }
 
 h1 {
-  margin-top: 0px;
+    margin-top: 0px;
 }
-
 
 /* these are all the same for now, but easy to modify individually */
 #container-ad, #container-error, #container-exp, #container-consent, #container-instructions, #container-questionnaire {
+    position: absolute;
+    top: 0px; /* Header Height */
+    bottom: 0px; /* Footer Height */
+    left: 0px;
+    right: 0px;
+    padding: 100px;
+    padding-top: 5%;
+    border: 18px solid #f3f3f3;
     background: white;
-    margin: 30px;
+}
+
+#container-consent, #container-exp {  /* override padding to fit long consent forms and exp display*/
+    padding: 10px;
+    padding-left: 20px;
+    padding-right: 20px;
+    padding-top: 1%;
 }
 
 /* ad.html  - the ad that people view first */
 #adlogo {
+    float: right;
     width: 140px;
     padding: 2px;
     border: 1px solid #ccc;
-    margin-right: 1em;
 }
 
 /* consent.html - the consent form */
 #consent {
-  margin: 0 auto;
-  margin-top: 0;
-  width: 800px;
-  font-size: 100%;
+    margin: 0 auto;
+    margin-top: 0;
+    width: 800px;
+    font-size: 100%;
 }
 
 .legal {
-  text-align: justify;
-  font-size: 11pt;
-  line-height: 1.1em;
+    text-align: justify;
+    font-size: 11pt;
+    line-height: 1.1em;
 }
 
 #trial {
-  margin: 0 auto;
-  width: 800px;
-  text-align: center;
+    margin: 0 auto;
+    width: 800px;
+    text-align: center;
 }
 
 #query {
-  color: darkgray;
+    color: darkgray;
 }
 
 .question {
-  margin-bottom: 30px;
+    margin-bottom: 30px;
 }
 
 /* errors */
 #error img {
-  padding-right: 10px;
-  padding-top: 15px;
+    padding-right: 10px;
+    padding-top: 15px;
 }
--- a/psiturk/experiment.py
+++ b/psiturk/experiment.py
@@ -276,12 +276,6 @@ def advertisement():
                 workerid=worker_id
             )
         else: 
-            # `thanks.html` was taken out in v2 with the introduction of the psiturk ad server
-            # but put back in in v2.1.2 because "local thanks page needed when reloading the 
-            # local ad and a worker has already completed the task." As far as I, @deargle, can interpret,
-            # a user, be they an AMT worker or not, should not typically reach this page unless 
-            # they're fooling around with URLs.
-            # 
             # Show them a thanks message and tell them to go away.
             return render_template( 'thanks.html' )
     elif already_in_db and not debug_mode:
--- a/psiturk/experiment.py
+++ b/psiturk/experiment.py
@@ -263,11 +263,11 @@ def advertisement():
         # them to start the task again.
         raise ExperimentError('already_started_exp_mturk')
     elif status == COMPLETED:
-        # They've finished the experiment but perhaps haven't submitted the HIT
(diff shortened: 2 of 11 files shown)


deargle:
@gureckis @pfeyz

jhamrick:
Woohoo this is awesome! I'll do a code review too to help out with this as I think this is pretty important. Maybe I will do a PR after this to include integration with something like Let's Encrypt :-)

psiturk/example/static/css/style.css
```diff
@@ -11,67 +11,80 @@ body {
...
 
 h1 {
-  margin-top: 0px;
+    margin-top: 0px;
 }
 
-
 /* these are all the same for now, but easy to modify individually */
 #container-ad, #container-error, #container-exp, #container-consent, #container-instructions, #container-questionnaire {
+    position: absolute;
```
jhamrick: These lines were removed on purpose, so adding them back in will introduce a regression -- see #159

deargle:
Code review would be great! The problem with Let's Encrypt is that they
won't issue certs for IP addresses
<https://community.letsencrypt.org/t/certificate-for-public-ip-without-domain-name/6082>
-- so domain name registration would still be required.

psiturk/experiment.py
```diff
@@ -263,16 +263,21 @@ def advertisement():
...
+            return render_template(
+                'thanks-mturksubmit.html',
+                using_sandbox=(mode == "sandbox"),
+                hitid=hit_id,
+                assignmentid=assignment_id,
+                workerid=worker_id
+            )
+        else: 
+            # then AMT isn't involved. Show them a thanks message and tell them to go away
+            return render_template( 'thanks.html' )
```
jhamrick: Doesn't this still need to have the template variables like `is_sandbox`, `hitid`, `assignmentid`, and `workerid`? Unless I'm missing something here? And I'm a bit confused about the comment -- isn't this used if we're relying on the psiTurk ad server, which does still involve AMT?

psiturk/psiturk_shell.py
```diff
@@ -941,60 +944,89 @@ def hit_create(self, numWorkers, reward, duration):
...
-                             '  template size uploadable to the Ad server is',
-                             '  1048576 bytes!' % size_of_ad])
-            return
+            # register with the ad server (psiturk.org/ad/register) using POST
+            if os.path.exists('templates/ad.html'):
+                ad_html = open('templates/ad.html').read()
+            else:
+                print '\n'.join(['*****************************',
+                                 '  Sorry, there was an error registering ad.',
+                                 '  Both ad.html is required to be in the templates folder',
```
jhamrick: Might be worthwhile to update this comment to be a little clearer -- "both" doesn't make much sense in this context. So maybe something like "Sorry, there was an error registering the ad. The file ad.html is required to be in the templates folder of your project so that the ad can be served!"

psiturk/psiturk_shell.py
```diff
@@ -941,60 +944,89 @@ def hit_create(self, numWorkers, reward, duration):
...
-        # what all do we need to send to server?
-        # 1. server
-        # 2. port
-        # 3. support_ie?
-        # 4. ad.html template
-        # 5. contact_email in case an error happens
+            size_of_ad = sys.getsizeof(ad_html)
+            if size_of_ad >= 1048576:
+                print '\n'.join(['*****************************',
+                                 '  Sorry, there was an error registering ad.',
```
jhamrick: Same thing here, maybe -- "Sorry, there was an error registering the ad..."

psiturk/psiturk_shell.py
```diff
@@ -941,60 +944,89 @@ def hit_create(self, numWorkers, reward, duration):
...
-        # register with the ad server (psiturk.org/ad/register) using POST
-        if os.path.exists('templates/ad.html'):
-            ad_html = open('templates/ad.html').read()
-        else:
-            print '\n'.join(['*****************************',
-                             '  Sorry, there was an error registering ad.',
-                             '  Both ad.html is required to be in the templates folder',
-                             '  of your project so that these Ad can be served!'])
-            return
+        if use_psiturk_ad_server:
```
jhamrick: Just a suggestion, but maybe create a separate function like `create_psiturk_ad` that takes takes care of creating the ad, and then call out to the function -- that will make the `hit_create` function a bit less bulky.

psiturk/experiment.py
deargle: @jhamrick I admit that I'm a bit confused about the purpose of `thanks.html`. It was [put back in 2.1.2](https://github.com/NYUCCL/psiTurk/commit/8195852822d17c902ab72608bebe28d7039d6f5a) by @gureckis. It doesn't use `is_sandbox`, `hitid`, `assignmentid`, or `workerid` anymore (it used to, before psiturk 2.0. It used to have the "click here to submit your hit" button). I couldn't figure out a way how workers would ever get to this page if they were using AMT. If they try to revisit the ad page, AMT just says "no more hits available to you in this group" and doesn't show the ad... at least that's how it went for me. Todd said this in the commit: "local thanks page needed when reloading the local ad and a worker has already completed the task." 

The only time I could see it being reached is if it were in debug (non-AMT) mode and a worker re-visited the /ad page. I guess though that /ad wouldn't be involved in debug mode.

My confusion aside, it should all still work fine though, just removing the "then AMT isn't involved" part.

psiturk/psiturk_shell.py
```diff
@@ -941,60 +944,89 @@ def hit_create(self, numWorkers, reward, duration):
...
+                'ad_html': ad_html,
+                # 'amt_hit_id': hitid, Don't know this yet
+                'organization_name': str(self.config.get('HIT Configuration', 'organization_name')),
+                'experiment_name': str(self.config.get('HIT Configuration', 'title')),
+                'contact_email_on_error': str(self.config.get('HIT Configuration', 'contact_email_on_error')),
+                'ad_group': str(self.config.get('HIT Configuration', 'ad_group')),
+                'keywords': str(self.config.get('HIT Configuration', 'psiturk_keywords'))
+            }
 
+            create_failed = False
```
jhamrick: Seems like there is a fair amount of code duplication here, especially with `hit_config`... it looks to me like the major differences are the `ad_location` and then the logic afterwards that checks if it failed or not, so I think it should be possible to at least share the code for the `hit_config` (maybe make a function to generate the hit config, and allow passing in the `ad_location` manually?)

psiturk/psiturk_shell.py
```diff
@@ -1054,17 +1080,19 @@ def hit_create(self, numWorkers, reward, duration):
                              '    ________________________',
                              '    Total: $%.2f' % total])
             if self.sandbox:
-                print('  Ad URL: https://sandbox.ad.psiturk.org/view/%s?assignmentId=debug%s&hitId=debug%s&workerId=debug%s'
-                      % (str(ad_id), str(self.random_id_generator()), str(self.random_id_generator()), str(self.random_id_generator())))
-                print "Note: This url cannot be used to run your full psiTurk experiment.  It is only for testing your ad."
+                if use_psiturk_ad_server:
```
jhamrick: It would be nice to still print the URL to the ad, just using the local server rather than the psiturk ad server.

jhamrick:
Ahh, good point. Bummer. Well, it might still be worthwhile to do if you are running psiTurk behind a domain name (which I usually am).

psiturk/example/static/css/style.css
deargle: @jhamrick Ah okay mb. Hmm I'll try to figure out a different way to make it match the 'push the complete button' page used when using the psiturk ad server.

psiturk/experiment.py
jhamrick: Ahh, ok, so those variables were accidentally left in even though they weren't being used. Got it. Looks good then!

jhamrick:
Oh one additional comment is that since you are adding in new config options, it would be good to add those to the relevant documentation files too:

* https://github.com/NYUCCL/psiTurk/blob/master/doc/config/server_parameters.rst
* https://github.com/NYUCCL/psiTurk/blob/master/doc/config/shell_parameters.rst
* https://github.com/NYUCCL/psiTurk/blob/master/doc/configuration.rst

deargle:
@jhamrick right, good idea. My original thought was to not document this until it had been tested a bit -- like the "experimental tunnel server" was never documented. But since things are going into the default config file users will wonder about them. I'll document it.

Also a github noob question -- do I have to do username mentions each time for you to know that I responded since you're not the thread author? Or would you get notification regardless? .

Thanks a ton for the code review.

jhamrick:
Ah, I see. I would probably err on the side of documenting it and maybe just adding a note to say it's an experimental feature rather than not documenting it at all.

> do I have to do username mentions each time for you to know that I responded since you're not the thread author? Or would you get notification regardless?

Once someone comments on a thread they get notifications by default, so it's not necessary to @-mention people. If they haven't commented yet, though, then @-mentioning them will notify them (unless they are already automatically subscribed to all issues on the repo, in which case they'll get notified regardless). The one thing that *doesn't* get notified is if you push a new commit to a PR that addresses someone's comments -- in that case you want to leave a comment saying you've made the changes.

No problem for the code review! Glad to help on the occasional chance that I have a few spare moments 😛

psiturk/example/static/css/style.css
deargle: Oh I get it now. the `style.css` on psiturk.org, which I incorporated into this commit, is outdated. I'll just revert all changes to this file.

psiturk/psiturk_shell.py
deargle: Yeah I had noticed that too, cringed at the grammar, then pressed on. You're right, this is a good time to fix it.

deargle:
@jhamrick code commits done, working on docs now. I'm going to pull all the commentary out of the default config file. After looking at it again, it may be overwhelming to non-technical newcomers. I'll move the commentary into the docs.

deargle:
Okay documentation pushed up now too. Phew.

jhamrick:
👍 looks great to me! Though note, I haven't actually tested it, just looked through the code.

GitHub emojis are the best! 🎉

psiturk/psiturk_shell.py
Portll: There's also the file size in the tastefully unique 'byes'. Worth fixing if
you're working in the file anyway, but I've avoided single figure
typo-commits because people get annoyed at them bloating the commit log.

~ John

deargle: ooo that's too many byes. I fixed it, but I'm afraid my rebase skills are
weak so in fixing this I created yet another commit. I imagine I'll be
squashing all this down to one commit after a green light though, so
probably not a big deal.

deargle:
Either things have changed a bit in nginx, or my gist was always wrong.
Regardless, try taking out `ssl on;` and adding `ssl` to the listen line,
like this: `listen 443 ssl;`.

Also, for good measure, specify the ssl protocols to accept in the nginx
server block, like in here:
http://nginx.org/en/docs/http/configuring_https_servers.html

hyiltiz:
Removed `ssl on;` and changed to `listen 442 ssl;` as suggested with `ssl_protocols TLSv1.2;` but I am getting the same error regardless of whether a) I use `proxy_pass https://...` and specify the SSL keys in `config.txt` for `psiturk`, or b) I use `proxy_pass http://...` but comment out the SSL keys in `config.txt`.

hyiltiz:
Yup, it was 443. Was a typo.

PR #228, opened 2016-04-30
this `atkill` handler does _not_ do what it claims. It claims to hand…

### LINKED ISSUE #227: psiturk -s script.sh shuts down server
See [this google group discussion](https://groups.google.com/d/topic/psiturk/CrhHY6z8QOA/discussion). 

IMO `psiturk -s` _shouldn't_ shut down the server. But it does because of [this code](https://github.com/NYUCCL/psiTurk/blob/d9fd9fe9c88c1ce3a9507b6f54eda49c21f1fb48/psiturk/psiturk_shell.py#L1840-L1850). A code comment in that exit handler says that it's intended to catch keyboard interrupts and exit cleanly, but actually, `atexit` only catches _normal_ program termination, _not_ keyboard interrupts. (See [here](https://pymotw.com/2/atexit/#when-are-atexit-functions-not-called)). If it were fixed, then a psiturk server process could be run in the background, and --script could be called on top of it. Would be really nice.

deargle:
…le keyboard interrupts but that is precisely what the module is designed to not do. what this codeblock _does_ do is kill the server whenever `psiturk -s` is invoked. That's bad. `psiturk -s` should shut down its own servers if that is what the user wants to do. other code already handles shutting down the server on invocation of `exit` or `server off`. this code simply has to go.

Closes #227

### DIFF
--- a/psiturk/psiturk_shell.py
+++ b/psiturk/psiturk_shell.py
@@ -15,7 +15,6 @@ import datetime
 import urllib
 import signal
 from fuzzywuzzy import process
-import atexit
 
 from cmd2 import Cmd
 from docopt import docopt, DocoptExit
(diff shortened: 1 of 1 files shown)


deargle:
:tada::confetti_ball:

PR #232, opened 2016-05-26
changing the error message for RDS connections for IAM users

deargle:
December 2015 AWS enabled IAM access control for Mturk. This commit
edits the error message to be DB-services specific. Right now it makes
users think that their credentials aren't good for anything, but
actually the message here is only triggered if RDB services can't be
accessed.

Other commits should maybe change other login messages too to be more
specific, but this is a start.

Sorry for the obscene length but at least the message is informative!

see [this group
thread](https://groups.google.com/forum/#!topic/psiturk/1nx8D-tsE74)


deargle:
I approve myself.

PR #233, opened 2016-06-08
use openshift db cartridge, if one is installed

deargle:
openshift will host mysql or postgresql relational DBs for you, if you
add those cartridges. this commit tells db.py to use the first installed
openshift db it finds. note that it is theoretically possible to have
both postgresql and mysql addon cartridges installed at the same time.
don't do that, that would be silly.

No editing config.txt required -- the location of the openshift db can
change, so relying on the env vars is really the best way to go. and
it's not easy to integrate the env vars into config.txt.

See README at https://github.com/deargle/openshift-psiturk-cartridge for
an example of getting mysql up and running.

### DIFF
--- a/psiturk/db.py
+++ b/psiturk/db.py
@@ -1,13 +1,18 @@
-
 from sqlalchemy import create_engine
 from sqlalchemy.orm import sessionmaker, scoped_session
 from sqlalchemy.ext.declarative import declarative_base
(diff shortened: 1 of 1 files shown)


deargle:
I have been successfully using this code in live apps for both , so I'm going to merge this.

PR #242, opened 2016-07-11
Update task_parameters.rst

lositsky:
Added detailed explanation of how experiment_code_version determines which participants in database are used for counterbalancing and for deciding which condition to assign an incoming participant.


deargle:
Thanks! I merged in a different PR that includes this commit.

PR #246, opened 2016-07-25
Reverse proxy support

Wildcarde:
There's not much changing here, the primary code is:

```
+        elif self.config.has_option('Server Parameters','revproxy'):
+            base_url = self.config.get('Server Parameters', 'revproxy') + "/ad"
```

This is added to two locations in the psiturk_shell.py file.  This enables you to supply an apache/nginx server address as the actual location for the psiturk server instead of relying on the locally generated high port number value. This proved critical for helping one of our users get their site serving out at a reasonable speed due to the inclusion of some extremely large resources in the package.  The functionality can be used by adding 'revproxy' to the 'Server Parameters' section of the config file and supplying a full http:// path to the base of the web app.

Inspecting the code and following the 'OPENSHIFT_SECRET_TOKEN' entries through the rest of the code it didn't appear that you would need to add this anywhere else, but I could be wrong as I'm not familiar with the full code base. As it stands now the code works in debug mode and will be getting tested in production shortly by the end user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <img src="https://psiturk.org/static/images/psiturk_logo_README.png">
-
+test
 [![Build Status](https://travis-ci.org/NYUCCL/psiTurk.png?branch=master)](https://travis-ci.org/NYUCCL/psiTurk)
 [![Coverage Status](http://coveralls.io/repos/NYUCCL/psiTurk/badge.png?branch=master)](http://coveralls.io/r/NYUCCL/psiTurk?branch=master)
 [![Code Health](https://landscape.io/github/NYUCCL/psiTurk/master/landscape.png)](https://landscape.io/github/NYUCCL/psiTurk/master)
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
 <img src="https://psiturk.org/static/images/psiturk_logo_README.png">
-test
 [![Build Status](https://travis-ci.org/NYUCCL/psiTurk.png?branch=master)](https://travis-ci.org/NYUCCL/psiTurk)
 [![Coverage Status](http://coveralls.io/repos/NYUCCL/psiTurk/badge.png?branch=master)](http://coveralls.io/r/NYUCCL/psiTurk?branch=master)
 [![Code Health](https://landscape.io/github/NYUCCL/psiTurk/master/landscape.png)](https://landscape.io/github/NYUCCL/psiTurk/master)
--- a/psiturk/psiturk_shell.py
+++ b/psiturk/psiturk_shell.py
@@ -142,11 +142,11 @@ class PsiturkShell(Cmd, object):
         database_url = self.config.get('Database Parameters', 'database_url')
         host = self.config.get('Server Parameters', 'host', 'localhost')
         if database_url[:6] != 'sqlite':
-            print("*** Error: config.txt option 'database_url' set to use " 
+            print("*** Error: config.txt option 'database_url' set to use "
                 "mysql://.  Please change this sqllite:// while in cabin mode.")
             quit_on_start = True
         if host != 'localhost':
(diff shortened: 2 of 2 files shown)


Wildcarde:
We've found an issue with this related to the ad.html file delivered to mturk. Specifically the {{server_location}} doesn't appear to get filled in correctly and still finds the <ip>:<port> variant of the URL instead of the fqdn port 80 address. For the time being we've replaced the ad.html entry with the correct URL and that seems to be working but ideally that would not be necessary.

deargle:
Are you already familiar with the new use_own_ad_server functionality? This allows for a reverse proxy server too, but the downside is that it requires that you host your own https cert. I agree that it would be nice to use a reverse proxy while still using the ad server.

> We've found an issue with this related to the ad.html file delivered to mturk. Specifically the {{server_location}} doesn't appear to get filled in correctly and still finds the : variant of the URL instead of the fqdn port 80 address. For the time being we've replaced the ad.html entry with the correct URL and that seems to be working but ideally that would not be necessary.

[Here's](https://github.com/NYUCCL/psiTurk/blob/75c078ef94a9d0ca4c02089ea6a273a57863de7d/psiturk/psiturk_shell.py#L1099-L1112) the relevant parts of the code that sends the ip;port up to the ad server: 

Also, if you're using openshift, [this cartridge](https://github.com/deargle/openshift-psiturk-cartridge) will give you an nginx server in front of psiturk automatically.

Wildcarde:
I hadn't looked at the non ad-server based setup, we host VMs locally for our end users on an as needed basis and haven't gotten automation to the point where we can easily farm them out an ssl cert yet.  We would (I believe) have to get general council and IRB approval to use openshift, which while possible isn't something we've pursued at this point. Typically we don't need to do much of anything other than allowing them a high number port to use and it all works as intended. This issue cropped up because of the end users need to deliver a unity web app as part of the experiment, which unfortunately gunicorn+flask isn't well suited for and generates a number of time out issues. This was simply a way to solve that particular problem without breaking the model overall if it could be avoided. I'll take a look at the additional code location listed here and see if I can resolve the rest of the issues with this approach.

deargle:
The ad gets the server location passed to it [right here](https://github.com/NYUCCL/psiTurk/blob/8464eaf3857efc431f33bb15b380223e85ced495/psiturk/psiturk_shell.py#L1108). This is where the revproxy would need to be inserted instead of the result of the `get_my_ip()` call. I'm mulling over the best way to integrate a `revproxy` config with the `use_own_ad_server` functionality.

deargle:
Here are the needs for the config file

### If using psiturk ad server:

when creating the ad, 

* the ad_location, hosted on ad.psiturk.org. is automatically determined and passed to mturk
* need to pass the externally-accessible server location. Currently, `get_my_ip()` is called and passed as the externally-accessible server location. But we want to be able to pass a revproxy. This is different from the `host` param in config.txt, because that determines what ip the gunicorn server binds to, and is typically 0.0.0.0 (not a fqdn). and a revproxy may be in front of gunicorn. **this will _not_ end in `/ad`**


### If hosting your own ad (not using the psiturk ad server):

* need to explicilty set ad_location, which is passed to mturk. **this _will_ end in `/ad` (or `/pub`)**
* no need to specify a separate reverse proxy, because `ad_location` accepts a fqdn, and it can be the revproxy location. urls coming from the ad will be relative to the original `ad_location`

So we only ever need to pass one config, but never more than one. I was thinking about making them the same config param, but that seems more confusing than necessary.

Thinking about naming the using-ad-server rev-proxy `adserver_revproxy`, so as to not confuse those using their own ad server. But that seems like a terrible name.

deargle:
Closing this, it has been implemented. See f66f3a664b35ac5adc3679f4059743748c000af2

PR #249, opened 2016-08-17
Adding a reference to the LICENSE

pmlandwehr:
I've gone ahead and made a `conda` package out of this repo for distribution via the community-curated `conda-forge` channel. (If y'all want to help maintain it, that would be cool, but is unrelated.) We've started included links to license files in the conda builds themselves, but doing so requires that the license be in the manifest itself. If this pull gets incorporated into the next pypi version, that would let us include said link.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,3 +4,4 @@ recursive-include psiturk/default_configs *
 recursive-include psiturk/psiturk_js *
 recursive-include psiturk/tunnel *
 include requirements.txt
+include LICENSE


deargle:
Any insights on the [two](https://groups.google.com/forum/#!searchin/psiturk/anaconda|sort:date/psiturk/V9vEc5Cmp_4/3DkncihNBAAJ) [issues](https://groups.google.com/forum/#!searchin/psiturk/anaconda|sort:date/psiturk/QVrh91_jzVc/WZCNjHFlBwAJ) that users have been having since you created the psiturk anaconda package?

pmlandwehr:
Hmm, interesting! In hindsight, I kept the requirements pretty loose when making the [feedstock](https://github.com/conda-forge/psiturk-feedstock) for conda-forge, so am going to use this plus the fact that the version has ticked up to get all of the version requirements a bit tighter. (If `requirements.txt` *can* be more loosely pinned, that would be great.) I'll also broaden out the number of imports we test at build, since I could have missed a `sqlalchemy` or `psutil` version issue by not fully testing those.

More generally - in [this](https://groups.google.com/forum/#!searchin/psiturk/anaconda%7Csort:date/psiturk/QVrh91_jzVc/WZCNjHFlBwAJ) issue the user is still installing the package via `pip` - they're just using conda to install pip first. If they want to install `psiturk` in via conda-forge, they should tell `conda` to check the conda-forge channel: `conda install -c conda-forge psiturk`

(Also, as I was typing this up you dropped an issue in the [feedstock](https://github.com/conda-forge/psiturk-feedstock), which is perfect. Please feel free to tell folks having conda install issues to put the problem in there.

deargle:
Sounds good. I tested psiturk using psutil 4.3.1 and it worked fine on debian linux. Anaconda 4 ships with psutil 4.3.0 so the feedstock can just use that if you want.

pmlandwehr:
Ah, boo. That's an issue, certainly - you might want to check out setting up some CIs to handle the alternate builds. (Not my forté in any way, but that's what the conda-forge admins have set up to test all of the different builds.) Anyhoo! I've switched to all of the pinned libraries that were available on `conda`, but for those that weren't I chose the oldest available.  In my [current pull](https://github.com/conda-forge/psiturk-feedstock/pull/3), I've switched to:
* `cmd2 ==0.6.8`
* `sh ==1.11`
* `fake-factory ==0.5.8`
* `fuzzywuzzy ==0.11.0`
* `gitpython ==2.0.8`
* `setproctitle ==1.1.10`
* `ua-parser ==0.7.1`
* `user-agents ==1.0.1`

deargle:
Not my forté either. Maybe one day...

I just dropped a note on my closed PR but I'll say it here too: I think `psutil` in the anaconda package needs to be allowed to be greater than 1.2.1, or anaconda explodes. For the package, the version could:

* be unspecified (which would default to using the anaconda-bundled 4.3.0, which works)
* be `>=1.2.1`, which would default to the anaconda-bundled 4.3.0

But the status quo of `==1.2.1` is breaking things for anaconda.

deargle:
Were you seeing errors in the build that went away when you switched down to the oldest-available versions of those packages? I think we'll still get some psiturkers who are using anaconda who will try to install psiturk using pip, so maybe I need to go in and bump our requirements.txt to use the same versions you found to work.

pmlandwehr:
I saw no errors, but my testing for `conda-forge` is *extremely* minimal - essentially, I test to make sure that all imports can be performed and all command line commands can be run without crashing. I *don't* run any of the additional module tests. My assumption was that they would - that the pinnings are themselves a bit strict. If you give the pinnings a test and they work, my rec. would be to tell users to just install via the `conda-forge` channel. That should manage all of the recs seamlessly.

PR #253, opened 2016-12-30
Bug Fix

jacob-lee:
The methods get_trial_data, get_question_data and get_eventdata were
not catching TypeError and ValueError exceptions as intended, causing data download in the psiturk shell to fail.

The list of exception classes needs to be enclosed in parentheses.


deargle:
Indeed. Thanks!

PR #256, opened 2017-01-30
Committing for testing. Bug Fixes and Improvements to Hit Creation

jacob-lee:
# Description of Changes
The following bugs and UX issues were noticed when creating a HIT
in the command line tool:

* Partial command input raises uncaught exception
    Example Input: hit create 9
    Now it queries for the remaining arguments.
* The regex to validate the reward per HIT fails on inputs like:
      -10.00
      3J55
      J55
   New stricter regex used, along with cast to float.
* Uncaught exception when duration can not be cast as a float, e.g.
   when receiving a non-numeric string as float.

# Notes

* The arguments passed to amt_services methods have been cast
to their numeric equivalents. My tests, and the documentation for
boto.mturk and amazon's amt API suggest this should be fine.

* Maybe the reward amount should be explicitly tested >= 0, though
I cannot think of a way that a negative value might pass through the regex filter.

### DIFF
--- a/psiturk/psiturk_shell.py
+++ b/psiturk/psiturk_shell.py
@@ -937,34 +937,46 @@ class PsiturkNetworkShell(PsiturkShell):
                 if user_input != 'y':
                     return
 
-        interactive = False
+        # Validate arguments
+        # numWorkers
         if numWorkers is None:
-            interactive = True
-            numWorkers = raw_input('number of participants? ')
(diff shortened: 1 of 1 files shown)


jacob-lee:
I think a confirmation prompt is a good idea. I can work on that.

deargle:
Tested, works. Thanks @jacob-lee!

jacob-lee:
Great!

Issue #244, opened 2016-07-11
PsiTurk using old number of conditions


deargle:
noticed by @lositsky in this thread: See https://groups.google.com/forum/#!topic/psiturk/5957-nhV1qE

> It turns out that inside of experiment.py, the code that determines which condition to assign reads through the database of participants and figures out how many participants were assigned to each condition in the database (line 134 - 153) that were run using the same code version (experiment_code_version). Since I previously ran participants with 20 conditions, it ends up using that number, even though the new number of conditions is 4. So it turns out that if I want to change the num_conds variable in config.txt, and still write participants to the same database, I also have to change the experiment_code_version. I didn't realize that experiment_code_version is the variable that determines which participants in the database are taken into account for counter-balancing.

The docs have been updated with a workaround, but it'd be nice if experiment.py got a bugfix.

deargle:
Thanks @ariekahn! The warning in the docs has been removed.

ytimber:
Hi @deargle, I'm wondering how should I change my task.js code after changing parameters in config.txt, so that my js code will be able to receive the counterbalanced condition name generated by psiturk? Thank you!

PR #257, opened 2017-02-06
always use current cond/counterbalance values from config

ariekahn:
Quick fix for #244 (PsiTurk using old number of conditions)


deargle:
Thanks!

PR #259, opened 2017-02-08
fixing the boto version in requirements.txt

braingineer:
boto 0.15 causes python 3 to crash (bad print statement)


deargle:
Heyo, so boto 2.45 is python 3 compatible? That's what was holding psiturk back once upon a time from upgrading itself.

braingineer:
@deargle - i haven't tried it yet with python 3, it was least stopping me from the `pip install -r requirements.txt`.  I'm running through and converting print statements right now... ;P

PR #260, opened 2017-02-08
changes to make the code python3 compliant (WIP)

braingineer:
hi all,

This isn't quite done. There's still something being thrown in gunicorn when I go to turn the server on. but it installs, and the shell does start, so that's a start.

Note: there is an issue with even the latest version of boto. Until it's fixed, one option is to go in and fix it. I've detailed the changes here: https://github.com/boto/boto/issues/3598 .  I hope to get to a PR for it soon. 

things changed:

1. all print statements have been made python3 
2. a couple exceptions were still in the old style
3. requirements.txt required some new versions 
4. maps and filters become iterators, so had to wrap some things with list
5. had to do some finnicky import stuff. mostly, urllib2 was restructured.  but also, in the use of __main__.py, experiment_server.py can be called as a non-packaged, which makes python throw errors for relative imports. I wasn't sure if it would always be imported in this way, so I put both version into a try-except.


requirements.txt
```diff
@@ -2,15 +2,15 @@ argparse==1.2.1
 Flask==0.10.1
 SQLAlchemy==0.8.3
 gunicorn==19.4.5
-boto==2.15.0
-cmd2==0.6.7
+boto==2.45.0
+cmd2==0.6.9
```
tleonhardt: I would strongly recommend moving to a much newer version of cmd2.  If moving to Python 3 only, I'd recommend 0.9.12 and for Python 2 and/or both Python 2 and 3 I'd recommend 0.8.9.    

But if you use `setup.py` you can set conditional dependencies differently based upon Python version.

deargle: Yeah I'm upgrading cmd2 as part of the creation of a testing suite.

tleonhardt: @deargle I'm one of the maintainers of `cmd2`.  If you run into any issues when upgrading, please feel free to create a Github Issue for us and me or one of the other devs would be happy to help.  But hopefully your experience is easy ;-)

deargle:
We're up-and-running with python3 support now

Issue #265, opened 2017-02-21
Example experiment: Bug in resubmit function?


ariekahn:
In the stroop example task.js, the questionnaire calls `psiTurk.completeHIT()` once the bonus computation is successful. If it fails though, `resubmit()`  tries to resend the data, and once successful, calls `finish()` instead, which looks like a holdover from an earlier version of the code.

Should `resubmit()` be calling psiTurk.completeHIT() as well?

[first try](https://github.com/NYUCCL/psiTurk/blob/a8345d1ad039f11d43ea8b13129c5f2cb8e8f28d/psiturk/example/static/js/task.js#L203)
```
psiTurk.computeBonus('compute_bonus', function() { psiTurk.completeHIT(); }); 
```

[resubmit](https://github.com/NYUCCL/psiTurk/blob/a8345d1ad039f11d43ea8b13129c5f2cb8e8f28d/psiturk/example/static/js/task.js#L188)
```
psiTurk.computeBonus('compute_bonus', function(){ finish() }); 
```

deargle:
Probably. There's some weird stuff in that whole "try again" bit. I suppose it just never fails the first time so that code is never reached, so no one has complained. Will patch though with your PR.

PR #266, opened 2017-02-21
replace finish() with completeHIT() in example task

ariekahn:
Quick fix for:
#265


deargle:
Looks like this was fixed in some other commit. Thanks though.

PR #273, opened 2017-03-26
Added confirmation dialog to hit_create()

jacob-lee:
* Implemented dialog confirmation as private method: _confirmDialog
* Asks for confirmation before HIT is created. created if response resolves to 'yes'. if 'no' or 'n', aborts creation of HIT.

### DIFF
--- a/psiturk/psiturk_shell.py
+++ b/psiturk/psiturk_shell.py
@@ -891,7 +891,7 @@ class PsiturkNetworkShell(PsiturkShell):
             self.live_hits = num_hits
 
     def hit_create(self, numWorkers, reward, duration):
-
+        ''' Create a HIT '''
         server_loc = str(self.config.get('Server Parameters', 'host'))
         inaccessible_but_do_it_anyways = False
         if not self.quiet:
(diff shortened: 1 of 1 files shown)


psiturk/psiturk_shell.py
```diff
@@ -988,10 +985,12 @@ def hit_create(self, numWorkers, reward, duration):
         if duration <= 0:
             print '*** duration must be greater than 0'
             return
-        
 
+        if not self._confirm_dialog('Create HIT? [y/n]: '):
```
deargle: needs to respect `self.quiet` (see the two confirmation dialogs earlier in this method call) so that `psiturk -e` can still be reliably used

deargle: I was envisioning some way to nicely display the options that will be used for the HIT if the user says 'y', perhaps putting [this code](https://github.com/jacob-lee/psiTurk/blob/8be8f6b8481e3e839afacee84ad333853d8fdd43/psiturk/psiturk_shell.py#L1049) into a function and repeating that, with some wording change to say something along the lines of "about to create a hit with the following settings"

jacob-lee: sure, I'll work on it.

jacob-lee:
Should check out now. Sorry for being taking so long, but got crushed at work.

deargle:
Thanks for your work on this. Sorry it took so long to merge.

PR #279, opened 2017-05-13
bump Flask and Werkzeug versions

alexanderrich:
Flask 0.12.1 allows range requests, allowing proper streaming and seeking for
videos.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 argparse==1.2.1
-Flask==0.10.1
+Flask==0.12.1
 SQLAlchemy==0.8.3
 gunicorn==19.4.5
 boto==2.15.0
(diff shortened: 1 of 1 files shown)


alexanderrich:
I've been trying to include long mp4 videos in an experiment, and have found that seeking within the video works buggily in Firefox and not at all in Chrome, and the videos don't even load in Safari. I think I got to the bottom of the issue, which is that until recently Flask didn't support byte range/partial content requests, which browsers use retrieving parts of large files. The recent Flask update adds that functionality.

Bumping flask and werkzeug doesn't seem to break anything, but It might be worth a few more people trying it before merging into master.

deargle:
Works well for me too, and all tests still pass. Thanks!

PR #283, opened 2017-09-06
Approving, bonusing, and server improvements

alexanderrich:
This branch contains several small features and improvements:

1. **Automatic bonus message**: adds a config option to write a bonus message that will be used whenever bonusing, so the psiturk user doesn't have to keep typing it. I chose "Thanks for participating!" as a default.

2. **Handling missing or duplicate submissions**: I sometimes encounter cases where I have 2 workers with status 4 for a single assignment. Often I get a message from one of the workers saying the MTurk signed them out automatically, or something of that sort. Ideally psiTurk would never set status to 4 without the HIT being sucessfully completed but as a quick fix I added code to make sure that if there are 2 submissions for an assignment, psiTurk checks which one is the true submission with mTurk and gives approval and bonuses properly. I also added code to make sure that when psiTurk is approving all assignments for a HIT, it **DOES NOT** automatically approve a participant if they are not present in the database, as this might mean they didn't really do the HIT.

3. **Allow multiple servers per user**: psiTurk now adds a hash of the current experiment directory to the server process name when starting the server. This allows the server to later be identified and distinguished from other psiTurk servers started by the user, allowing multiple servers to run simultaneously (as long as they're on different ports). (There is an old abandoned pull request (#198) regarding this issue, but I believe that solution could falsely tell the user their server is running when a different psiTurk server is using the same port.)

All three of these improvements have been tested by me on live HITs, though more testing is of course appreciated!

### DIFF
--- a/doc/config/hit_configuration.rst
+++ b/doc/config/hit_configuration.rst
@@ -19,7 +19,7 @@ like this:
 	psiturk_keywords = stroop
 	organization_name = New Great University
 	browser_exclude_rule = MSIE, mobile, tablet
-    allow_repeats = false
+	allow_repeats = false
 
 
 `title` [string]
--- a/doc/config/shell_parameters.rst
+++ b/doc/config/shell_parameters.rst
@@ -8,6 +8,9 @@ the psiturk shell.
 
 	[Shell Parameters]
 	launch_in_sandbox_mode = true
+	bonus_message = "Thanks for participating!"
+	use_psiturk_ad_server = true
+	ad_location = false
 
 `launch_in_sandbox_mode` [ true | false]
 -----------------------------------------
@@ -20,6 +23,14 @@ to `true` to lessen the chance of accidentally posting a live HIT to mTurk.
 
    `Overview of the command-line interface <../command_line_overview.html>`__
    	  The basic features of the **psiTurk** command line.
+
+`bonus_message` [string]
+-------------------------
+
+If set to a string, automatically uses this string as the message to
+participants when bonusing them for an assignment. If not set, you will be
+prompted to type in a message each time you bonus participants. (This message is
+required by AMT.)
       
 `use_psiturk_ad_server` [true | false]
 ---------------------------------------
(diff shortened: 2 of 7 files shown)


alexanderrich:
I think this is why the Travis build failed, but I'm not really a Travis expert/ don't know how to log in and change the build: https://github.com/travis-ci/travis-ci/issues/7940

deargle:
I just fixed the travis issue on master, thanks for the link. But I'm not sure how to retrigger your integration tests without you adding another commit.

These features sound cool! I'll test them the next time I'm running an experiment, which might be a while.

> Ideally psiTurk would never set status to 4 without the HIT being sucessfully completed

Sigh, yes, that would be wonderful. But MTurk won't allow it. The final POST submission to mturk _has_ to be done from the worker's browser; we can't do it for them via an ajax call or whatever. So the last chance we have to set status=4, is right before we submit the form to mturk, which is what we do currently. I can't find any way to catch whether that form submission fails 😢

alexanderrich:
Great, that looks like that fixed the Travis build! And I'll continue to stress-test these changes over the next few weeks.

deargle:
Can you write some documentation for the new bonus config?

alexanderrich:
Yep! Just updated the readthedocs and also added a commit with a couple small changes. Otherwise points (1) (2) and (3) from the original PR message have all been working well for me the last couple months!

deargle:
Thanks for adding documentation for my own stuff while you were at it :)

alexanderrich:
Interesting, I do get that every so often (I recently ran an n=400 study so the rare events became more salient). Is there any reason not to include status 3? It seems that if we have them as completed and MTurk says they submitted the mistake must be on our end (however it might be happening...)

deargle:
No reason not to include 3, I'm just at a loss for how it would happen.
Interesting that you've seen it.

alexanderrich:
Yeah, in an ideal world I would really dive into experiment.py to investigate. Might have to save that for the big refactor/switch to Python 3 some day...

deargle:
I've been using your code for a while now, with great success. I've tested the 1. automatic bonus message and 3. the dual servers. The missing and duplicate submission code is trickier to test, and I'm too lazy to write more unit tests (I'm not even sure how we _would_ write unit tests for this since it relies on mturk interaction), but just looking at the code, it should all work. I really want to pull this in to master for 1. and 3., even though 2. is not totally proven.

deargle:
By the way, thanks @alexanderrich! 🎉

PR #300, opened 2018-02-10
Changed psiturk_config.py to use os.path.join and code clean up.

jacob-lee:
Two changes to psiturk_config.py were implemented in this commit.

* In some places, config file paths were constructed using string concatenation. This was changed to use os.path.join.
* Code cleanup, e.g. lines < 80 chars.


gureckis:
:+1:

PR #304, opened 2018-03-18
remove unnecessary imports in psiturk_shell.py

fredcallaway:


deargle:
Note to self -- also remove `sqlalchemy` import

deargle:
PS, thanks @fredcallaway

PR #318, opened 2018-04-22
bf: reference to tally_hits should be update_hit_tally

ariekahn:
Status command is broken - looks like tally_hits was moved out of the class and a reference to it didn't get fixed.

Updated it to call update_hit_tally instead, seems to be working fine.


deargle:
Thanks. I didnt even know that functionality existed. Although with a dirty
hack I did a few days ago, calls to `psiturk status` won't work. Will fix
soon.

ariekahn:
Hah, neither did I. Called it by mistake and found it was broken, but the fix looked pretty simple.

deargle:
:+1:

PR #320, opened 2018-06-29
Get rid of the mysql-python dependency

### LINKED ISSUE #97: mysql-python dependency
currently we do not require mysql-python to install psiturk.  the reason is that mysql-python can be hard to install on some system (compilation errors are common).  however, if a user tries to use a mysql database psiturk will crash.  instead it should check if the library is installed and give a more helpful warning message that this feature has additional dependencies.

anselmrothe:
Until now, it was necessary to manually install the `mysql-python` package if one wanted to save the data in a [MySQL database](https://psiturk.readthedocs.io/en/latest/configure_databases.html?highlight=sql#using-a-self-hosted-mysql-database-recommended). Unfortunately, getting `mysql-python` properly installed was a nightmare. And `mysql-python` has hasn't been updated since 2014.

`pymysql` is a pure python alternative, which is [actively maintained](https://github.com/PyMySQL/PyMySQL) and which can be installed without problems. With `pymysql`, psiturk and your MySQL database can be happy again.

(Fixes #97)

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -218,7 +218,7 @@ be done at any time), further packages are needed.  On a Debian system, they are
 
 ::
 
-    aptitude install python-mysqldb python-mysqldb-dbg python-sqlalchemy libmysqlclient-dev
+    aptitude install python-pymysql python-sqlalchemy libmysqlclient-dev
 
 If you have additional specific issues, or if you can report the steps
 needed to install psiTurk on a particular Linux distribution, please help
--- a/doc/retrieving.rst
+++ b/doc/retrieving.rst
@@ -20,7 +20,9 @@ While the ``download_datafiles`` shell command is the simplest way to retrieve
 experiment data, a more powerful and flexible solution is to retrieve the data
 programmatically. Many languages offer libraries for interfacing with mysql and
 sqlite databases - below is an example using python and the sqlalchemy package
-to retrieve data from a mysql database. By including code such as this at the
+to retrieve data from a mysql database. We add `+pymysql` to the `db_url` to let
+sqlalchemy make use of pymysql package. (You can leave the database_url in config.txt
+as `mysql://` though -- psiturk adds `+pymysql` internally). By including code such as this at the
 beginning of your analysis script, you can be sure the the data you're analyzing is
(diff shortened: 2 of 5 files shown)


gureckis:
nice @anselmrothe: do we want to may `pymysql` a requirement or optional install (as was `mysql-python`)?  it seems wise maybe to load as needed that was if mysql ever becomes a challenge again it doesn't stop new users (one of the advantages of the old approach)

gureckis:
also this change requires a few modifications to the docs that reference the older mysql library (e.g., https://psiturk.readthedocs.io/en/latest/install.html?highlight=mysql)

anselmrothe:
As far as I can tell, `pymysql` is a pretty safe bet (not more likely to become a reliability than the other packages in requirements.txt). The installation difficulties with `mysql-python` did not result from dealing with MySQL (which both packages do) but resulted from some C compilation difficulties which `pymysql` [got rid of](https://code.google.com/archive/p/pymysql/wikis/Goals.wiki).

So I'd have a slight preference for making `pymysql` a requirement but I let you make the judgement call here.

For the docs, I already made a search and there's only the part about [Debian systems](https://psiturk.readthedocs.io/en/latest/install.html?highlight=mysql#linux), where somebody who knows about Debian would need to check which packages need to be changed (I assume `python-mysqldb-dbg`).

Everything else in the docs can remain the same if we keep the pull request as it currently is. If we make `pymysql` optional, then we might want to update one or two places.

deargle:
It's just kind of bloat to require it. I don't use mysql when running on
heroku, but I do use postgres.

anselmrothe:
Make sense. I removed it from the requirements.

The new version checks whether `pymysql` is installed and prints a message otherwise (as for `mysql-python` before).

SQLAlchemy requires the database_url to begin with `mysql+pymysql://`. For backwards compatibility and convenience, the `+pymysql` gets added internally. So people don't have to update their existing config.txt files. The only other place beyond `db.py` that needed to be adjusted is `amt_services_wrapper.py`, which I updated.

The documentation is also updated.

gureckis:
looks like a worthwhile change, thanks @anselmrothe!

PR #329, opened 2018-08-15
enh: add parameter to control gunicorn server timeout

### LINKED ISSUE #285: increase gunicorn timeout to avoid worker timeout
I have a rather complicated experiment where I've written multiple custom routes that happen between blocks in the experiment.  One of the processes can sometimes take > 30 seconds to return an answer, and I'm running into an issue where the psiturk server seems to stop responding when this happens (at least, I think this is the issue).  The error returned in the server log is: 

`[28] [CRITICAL] WORKER TIMEOUT (pid:32)`

I dug into the gunicorn documentation a little bit, and found that its possible to manually change the timeout duration with the timeout flag ([link](http://docs.gunicorn.org/en/stable/settings.html#timeout))

To expose this argument, I was thinking about modifying [these lines](https://github.com/NYUCCL/psiTurk/blob/master/psiturk/experiment_server.py#L67) of the  `experiment_server.py` file by adding the timeout flag like this:

```
        self.user_options = {
            'timeout': config.get("Server Parameters", "timeout"), ## <- add this line
            'bind': config.get("Server Parameters", "host") + ":" + config.get("Server Parameters", "port"),
            'workers': workers,
            'loglevels': self.loglevels,
            'loglevel': self.loglevels[config.getint("Server Parameters", "loglevel")],
            # 'accesslog': config.get("Server Parameters", "logfile"),
            'errorlog': config.get("Server Parameters", "logfile"),
            'proc_name': 'psiturk_experiment_server',
            'limit_request_line': '0',
            'on_exit': on_exit
        }
```

I was wondering if someone with more expertise with psiturk might be able to tell me if this is reasonable, or a terrible idea.  Or is it already possible to modify the gunicorn `timeout` parameter within psiturk, and I just missed it?  Any advice would be greatly appreciated (and let me know if you need additional info)!

ariekahn:
Fixes #285

### DIFF
--- a/doc/config/server_parameters.rst
+++ b/doc/config/server_parameters.rst
@@ -21,6 +21,7 @@ command line.  An example looks like this:
     #keyfile = <path_to.key>
     #adserver_revproxy_host = www.location.of.your.revproxy.sans.protocol.com
     #adserver_revproxy_port = 80
+    #server_timeout = 30
 
 
 `host` [string]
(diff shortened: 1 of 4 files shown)


deargle:
Thanks!

PR #335, opened 2018-12-28
later SQAlchamy is compatible with mysql 8.0

hyiltiz:
Passed test at 9806225fa3a5dbb2c64c86bc42067a77462f9d72 with `mysql --version
mysql  Ver 8.0.12 for osx10.14 on x86_64 (Homebrew)` and `SQAlchemy 1.2.15`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 argparse==1.2.1
 Flask==0.12.2
-SQLAlchemy==1.1.18
+SQLAlchemy>=1.1.18
 gunicorn==19.4.5
 boto==2.15.0
 cmd2==0.6.7


deargle:
Thanks!

hyiltiz:
Also tested in Debian Testing and it works. To test, I installed mariadb, python2, build-essential, git, python-setuptools via `apt`, then installed the `master` branch via `python2 setup.py install` as root. Psiturk worked with `mariadb  Ver 15.1 Distrib 10.1.37-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2`.

hyiltiz:
At some point (I guess around when the next release will be created), the documentation should be updated s.t. people who mostly rely on psiturk website can still get things work. I had to bump into too many walls before I got a working setup. Some of those were due to old mysql packages in macOS and multiple mysql and openssl packages installed through Homebrew which screwed up the dependencies. None of those are related directly to psiTurk, but are all essential for it to work. While I hesitate to say this, maybe a Docker image would actually suit this so people can get a working environment in a single click.

deargle:
I think a few have created docker containers in the past, floating around
here and/or on the google group. And fair point that the documentation is
thin.

I know nothing about running all this on homebrew and mac etc. Had kind of
hoped that virtualenvs would have sufficed before a docker container was
necessary, but yeah openssl won't apply there. Probably that is why reports
have been popping up about ssl problems for mac users -- which I cannot
replicate because I am mac-less. I kind of head-scratch at people running
psiturk on mac though, because I don't envision hosting the live server on
mac at go-time. Maybe mac for a dev environment though. It all just works
for me in my linux mint dev VM, so again, I haven't bumped into the
problems. For running live, I use heroku, where things also "just work"
with openssl etc.

hyiltiz:
Agreed, getting a server up and running in any decent vanilla GNU/Linux is a breeze, but quite a few labs have a iMac or Mac Pro (aka the trash can) as a "server". We wouldn't advise it but when they already got one, it is easier to provide provide a working tutorial/documentation than to persuade them to buy something more.

I have never tried Heroku before, but I could get it work in vanilla Debian which is available in any VPS provider, among which is DigitalOcean which comes with $50 credit if you apply for the Github Student Pack. About VMs though, running a server inside it sounds a bit hairy.

deargle:
I put up some docs on using heroku for psiturk somewhere or other. It's
free, or impossibly cheap (<$5). At least, my usage of it has never
surpassed $5 even for >1k participants over a week span. Heroku is built on
top of aws. Shared resources, smallest unit you can control is a _project_
(not a server), and the project is tied to a git repo. You make all edits
locally and then push -- no editing directly on a heroku node (aka "dyno").
You push to your heroku repo, and it autodeploys. Gives you free https, rev
proxy server, monitoring, postgresql db, all with minimal config. At least,
monumentally easier config than would be for using an in-house mac server.
Abstracts away the server administration stuff.

So I develop in a linux vm, including running the psiturk server and an
unnecessary but nice nginx rev proxy, and when it's ready, I push from the
vm to the heroku repo, and create hits from the vm too. I use pageant for
ssh key management and i use putty to forward my agent to the vm so that I
can push and pull easily where I please.

Why do you think running a server within a vm would be "hairy"? I'm pretty
well versed in server administration, at least relative to the typical
psiturk user, so if there is something intimidating about it, I could write
a small doc guide to help get started. I ssh in to run shell commands,
winscp with notepad++ for editing files, and run host-only adapter with
virtualbox plus avahi-daemon plus bonjour servics on windows so that I can
access the psiturk dev server by hostname from my windows web browser. I
never use the vm gui.

hyiltiz:
I mentioned vm may get hairy because of the *overhead* (not sysadmin+maintenance), but that depends on the type of vm and configurations. That overhead may not matter for most psych experiments ran on mTurk though.

PR #338, opened 2019-02-20
Fixes 377 by parsing the json returned from httpbin.org.

jpalat:
This method is safe if the previous output is provided.


deargle:
Thanks!

PR #368, opened 2019-07-31
Add optional arguments to the debug command.

jacob-lee:
I've added the ability to supply optional arguments to the debug command to set workerid, assignmentid, and hitid. For example, in a multiple HIT scenario, one might need to parameterize a worker's task by data stored from their previous HITs for you. The current setting prevents that by using random strings for each of these.

### DIFF
--- a/psiturk/psiturk_shell.py
+++ b/psiturk/psiturk_shell.py
@@ -454,7 +454,7 @@ class PsiturkShell(Cmd, object):
     @docopt_cmd
     def do_debug(self, arg):
         """
-        Usage: debug [options]
+        Usage: debug [--workerid=workerid] [--hitid=hitid] [--assignmentid=assignmentid] [options]
 
         -p, --print-only        just provides the URL, doesn't attempt to
                                 launch browser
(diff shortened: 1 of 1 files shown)


deargle:
Wait, why use `debug` command at all if crafting and handing out custom
urls for participants? Why not just give them a hand-crafted url?

jacob-lee:
I'm not sure I understand. `debug` is a convenience command for debugging applications on a local server. One could always modify a URL by hand for that purpose, of course, but then why bother with the `debug` convenience command at all?

deargle:
Oh, I thought you were using this to craft urls for live studies. But
instead, you want to debug-test worker-repeats, yeah? Which would mean
holding workerid constant but allowing hitid and assignmentid to vary.

In a live scenario, a worker can never take more than one assignment from
the same hit (mturk prevents it, in my experience).

Is there a use case for specifying more than just the workerid (the
assignmentid and hitid) to the debug command?

Another enhancement (just thinking out loud) is to allow a "repeat-worker"
flag which would accept a count, and would provide that many urls all with
the same workerid.

Separately, we could also build something into default.html to allow
specifying a workerid for repeat workers.

jacob-lee:
--workerid is the only option I've used, and its the most obviously useful. I added the others here for completeness, and because it was easy.

The only other option that I might consider is explicitly passing a condition parameter (note: I have not yet reviewed the changes you're proposing to handle custom condition assignment schemes).

deargle:
I mulled today that modifying complete.html to have a "do this task again"
button may be the most straightforward approach. I doubt noobs doing
repeats would readily think to use the debug command to get new urls -- as
for myself, i never use the debug command now that default.html includes
randomized url functionality. What do you think about the complete.html
approach? Javascript could do it easily since I think the workerid is
already passed to that url.

deargle:
Like this: https://github.com/NYUCCL/psiTurk/commit/5e8a9593562344696ff00fc604a965142a82f782

jacob-lee:
Its just a convenience command to construct a url and launch the browser. What you are arguing for is the removal of the debug command. That's your call. Different people, with different levels of engagement with this project, likely use it differently. For example, I was unaware of changes to default.html in newer versions of psiturk.

I am not sure that a "do this task again" button would be satisfactory for my use cases. For example, it assumes that the repeat uses the same task, but that is not the case for some of the longitudinal studies we've done.

deargle:
> For example, it assumes that the repeat uses the same task, but that is not the case for some of the longitudinal studies we've done.

My interest is piqued, I haven't thought much about longitudinal. Hmm so you have different tasks (experiment codebases) all saving to the same database? If so, how do you differentiate `task_1` from `task_n` for a given participant?

I can merge in this pr if it'd help you, even though I don't fully understand the use case. Would you be able to throw in a unit test or two for the different parameters into `test/test_psiturk_shell.py`? You could add a basic test like the following:

```
do_debug_commands = [
    (['debug -p --workerid abcd'], 'somepartoftheurlthatshouldbethereforsure'),
    (['debug -p --hitid 1234 --workerid abcd'], 'anotherthingtochecktheexistenceof')
]

@pytest.mark.parametrize('cmds,expected', do_debug_commands)
def test_do_debug(get_shell, cmds, expected, capsys):
    response = shell.runcmds_plus_hooks(cmds)
    captured = capsys.readouterr()
    assert expected in captured.out
```

PR #376, opened 2019-08-13
Quickstart documentation update

Speldosa:
First, I updated the requirement from Python 2.7 or lower to Python 3.0 or higher (since that's required now, right?). I don't know if there are restrictions on which Python 3 versions that can be used though.

I also added instructions of how to install a release since this was something that wasn't specified in the readme already.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ cffi==1.12.3
 chardet==3.0.4
 ciso8601==2.1.1
 Click==7.0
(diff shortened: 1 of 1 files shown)


deargle:
python 3 will be required for the psiturk v3 (master branch), but python2 still works for psiturk v2.3.x (python2 branch).

deargle:
readme.md as of right now has instructions for local installs. thanks for this PR though.

PR #388, opened 2020-01-17
changing dates to utc

dhalpern:
date times for beginhit/endhit currently use whatever timezone the server is in. maybe better to use UTC?

### DIFF
--- a/doc/recording.rst
+++ b/doc/recording.rst
@@ -24,7 +24,7 @@ add a line of data to the log, use ``psiturk.recordTrialData``:
 
 The list of values that you supply to ``recordTrialData`` will then be
 appended to the log. It is up to you how to structure those lists; you
-will have to parse them as part of your analysis. Each time you call ``psiturk.recordTrialData``, it will also record the time it was called (with a UTC timezone).
+will have to parse them as part of your analysis.
 
 .. _recording-unstructured-data-label:
 
--- a/doc/recording.rst
+++ b/doc/recording.rst
@@ -24,7 +24,7 @@ add a line of data to the log, use ``psiturk.recordTrialData``:
 
 The list of values that you supply to ``recordTrialData`` will then be
 appended to the log. It is up to you how to structure those lists; you
-will have to parse them as part of your analysis.
+will have to parse them as part of your analysis. Each time you call ``psiturk.recordTrialData``, it will also record the time it was called (with a UTC timezone).
 
 .. _recording-unstructured-data-label:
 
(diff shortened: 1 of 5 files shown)


deargle:
Probably should be fine. Will need to check how timezone affects usage of boto3

deargle:
Did you run unit tests with this PR, or if something was breaking using not-utc, can you write a test demonstrating the break?

dhalpern:
Nothing was breaking inside psiturk (as far as I know), it was more for consistency since timestamps inside datastring are in UTC (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime). In our particular use case, we are running an app in the cloud, potentially in a different timezone, so using UTC makes it easier to standardize. I imagine this might be useful for others doing similar things.

I haven't run any unit tests that would check both compatibility though (unless it's part of the Travis CI thing). I'm not super experienced with this, what would I need to do to add a unit test?

deargle:
Tests are stored in `tests/` and can be run using `pytest` (`pip install pytest`). E.g., I'm working on one right now for loading a project-specific `.psiturkconfig` file. I'm calling the test `test_project_psiturk_configfile` and I'm storing it in `test_psiturk.py`. To run just that test using pytest, I call, from the `psiTurk` package directory, `pytest tests/test_psiturk.py::test_proejct_psiturk_configfile`. 

Unit testing has a learning curve, but it's a great asset.

dhalpern:
Sorry for taking so long to get back to this, was dealing with a couple other things. So should I resubmit this pull request along with a new test? I'm sort of unclear what kind of test would be good here for such a simple change.

dhalpern:
Also my apologies for cluttering this up with a separate commit to my personal fork, still sort of learning about this pull request thing...

gureckis:
in this instance I'm not sure there is a test because the time is just going to be the time in whatever time zone you specify.  my only comment is that there are some calls to `.now()` that haven't been changed... why did you skip some?  also it would be nice to keep the docs up to date if we add a note about this someplace.  perhaps a line or two could be added to https://psiturk.readthedocs.io/en/latest/recording.html and https://psiturk.readthedocs.io/en/latest/retrieving.html just to mention that all time zones are in UTC?  

The only issue is if someone upgrades psiturk in the middle of a run and the time zones become misaligned or something.  But generally I agree that the localized time zone is problematic for secondary analyses or interfacing.

dhalpern:
Yeah I guess I only wanted to touch the ones that would change beginhit and endhit in the database just to be safe. The other instances of datetime.now() have to do with posting or expiring a HIT and I wasn't sure how that interacted with AMT services. There is a weird mix of .now() and .utcnow() in there currently that should maybe be cleaned up at some point though (utcnow is not recommended because it doesn't create the timezone correctly, see warning: https://docs.python.org/3.9/library/datetime.html). I'll update the docs to indicate the timezone recording.

dhalpern:
Ok docs are edited.

deargle:
Whatever the amt api is for expiry date time, I've hit and fixed at least
one bug with it before due to timezones borking things.

dhalpern:
I created a new pull request for this in order to follow the proper GitHub flow

PR #392, opened 2020-01-23
Changing dates to UTC

dhalpern:
same as previous pull request but using a different branch to follow GitHub flow

### DIFF
--- a/psiturk/api/__init__.py
+++ b/psiturk/api/__init__.py
@@ -237,7 +237,7 @@ class TaskList(Resource):
                 func=do_approve_all,
                 trigger='interval',
                 minutes=max(int(int(data['interval']) * 60), 30),
-                next_run_time=datetime.now()
+                next_run_time=datetime.datetime.now(datetime.timezone.utc)
             )
             return job, 201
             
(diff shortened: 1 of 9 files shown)


deargle:
continuing discussion from https://github.com/NYUCCL/psiTurk/pull/388

I think we should standardize switch all to `datetime.now(timezone.utc)` . I wrote the "campaign" stuff, and there's tests for that. As long as the campaign unit tests pass, it's okay to change it.

dhalpern:
I changed all the date times that I could find to UTC and it seems to have passed all tests.

psiturk/api/__init__.py
```diff
@@ -237,7 +237,7 @@ def post(self):
                 func=do_approve_all,
                 trigger='interval',
                 minutes=max(int(int(data['interval']) * 60), 30),
-                next_run_time=datetime.now()
+                next_run_time=datetime.datetime.now(datetime.timezone.utc)
```
deargle: Incorrect, should not be a second datetime here

deargle: I'm working on this. Turns out that the APScheduler needs special timezone treatment. I'll push to your PR.

deargle:
Merged. Now the big problem, of course, is that readthedocs docs still point to the master branch, even though this is a development branch. So the docs say that utc is used, but it's actually not, for everyone on pip. Still need to reconfigure readthedocs to point to tagged releases somehow.

deargle:
Also, reminder to self, next time we should `git squash` all these intermediary commits before merging

Issue #391, opened 2020-01-22
Why always the need to set DATABASE_URL env var when running on heroku?


dhalpern:
Currently, `psiturk_config.py` looks for `os.environ['DATABASE_URL']` if using Heroku and crashes if it can't find it. This variable doesn't get created unless you create a database when setting up your project which you may not realize if you plan on using your own db. The error thrown is fairly cryptic (just something like DATABASE_URL key can't be found) so it took me a while to figure out what was going on. I think ideally this would just throw a warning (e.g. "you haven't created a Heroku Postgres database for this project, please call heroku addons:create heroku-postgresql unless you intend on using your own db") rather than crash the whole thing?

deargle:
That only gets checked if the local `ON_HEROKU` is set. That shouldn't be set on the local development environment. It's intended to be an env var signal for if psiturk is actually running on heroku. https://github.com/NYUCCL/psiTurk/blob/578aa1950762d3b220da3a1036ecdeb715e7fca5/psiturk/psiturk_config.py#L71-L74

If you're saying it crashes on heroku, then yeah it'd be better to have a more descriptive error message. But the default sqlite _must not_ be used on heroku because dynos are not persistent. A persistent database _must_ be set. 

I suppose that psiturk _could_ rely on the database url set in config.txt, instead of using an env var as long as it wasn't sqlite or localhost. Ostensibly someone would want to override the database url set by heroku, I guess. Something like prefer env var first for DATABASE_URL, and failing that, check if an okay databaseurl set in config.txt. All (most?) of the config.txt settings should be overridable via env var.

On a bigger level, psiturk needs more user-friendly heroku integration.

dhalpern:
Yeah, I had ON_HEROKU set and was trying to use a preexisting mysql table so it felt like a case where it shouldn't throw an error. The solution of checking env vars first and then checking the config file sounds good to me, can try to make a pull request at some point.

dhalpern:
Actually, maybe I didn't understand this before but it actually uses the heroku database even if the database in config is different. It seems to me that the one in config should be prioritized...

deargle:
As an example of best practice, boto3 prefers env vars over config file:
https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html

Also, with pull requests, please follow the GitHub Flow, where you create a
separate branch off of master on your fork for each PR

dhalpern:
[...]

PR #393, opened 2020-01-23
Use alternative databases with Heroku

dhalpern:
This adds a parameter to the config file "use_heroku_db". If false, the database_url is whatever is set in the config file. If true, when on Heroku, the database_url is the one from env vars, following Heroku best practices.

This has the added benefit that Heroku no longer crashes if you do not create a PostGRES database (if you don't intend to use it).

This commit prioritizes the database in config file. I think this is probably the right way to do it due to what I said in #391 but if people think that's wrong, happy to try an alternative approach


dhalpern:
Also happy to add a test but not sure exactly what would be appropriate here.

PR #396, opened 2020-01-27
Heroku DB

dhalpern:
Deal with databases along the lines of the solution suggested in #391 . Checks for environment vars, then uses config if none. If ON_HEROKU is true and database_url contains 'sqlite' or 'localhost', throw an error


deargle:
This is nice, but with this change, I'd rather move to a more generalized option of removing the dependency on `ON_HEROKU` entirely. Or rather, letting the database_url be set via env var regardless of whether `ON_HEROKU` is set. But perhaps still retaining something like "ON_CLOUD" so that (1) we can guide users towards using that and throw your ephemeraldatabase if that's set. There's a few places sprinkled in the code where ON_HEROKU would need to be removed / changed if possible to ON_CLOUD.

And perhaps generalizing many more things to be env-var settable -- but maybe we just do the database bit for now, and merge this PR. Will come back to this again soon.

dhalpern:
I'm happy to go through and change all the occurrences of ON_HEROKU to ON_CLOUD, seems more general. 
And I feel like it would be would be good to start with this for now (and add other env-vars later).
It fixes the bad behavior where Heroku crashes without a clear error if you don't create a PostGRES db/set the DATABASE_URL env-var.

dhalpern:
Changed all ON_HEROKU variables to ON_CLOUD. Anything else that needs to get done to accept the PR?

dhalpern:
@deargle is there anything else I should do to get this merged in?

deargle:
I'll swing back to test this out the next time I'm developing on psiturk -- probably this week. I work on psiturk in waves, with other class and research todos etc.

deargle:
I'm going to merge this, but note to self that I still want to make reading from env vars more flexible across more (all?) config options. Thanks for your work on this.

Issue #406, opened 2020-03-17
Hit dispose is not a valid shell command


afdreher:
Commit 83f898dbc2e615ef5af21bdafb27af6054fba137 removed support for "dispose" as a synonym for "delete", but the psiTurk shell autocompletes it and the hit.txt help file references it.

deargle:
Resolved by #407

PR #407, opened 2020-03-17
Removes references to dispose

afdreher:
Short commit to removing references to the outdated command dispose #406


deargle:
Thanks!

PR #409, opened 2020-03-17
Creates basic help for config

### LINKED ISSUE #408: Help for config is missing / erroneous
The help for the shell command `config` is missing the necessary text file in the shell_help directory, and typing the command `config help` calls `self.help_server()` instead of `self.help_config()`, which requires the aforementioned .txt file.

afdreher:
This PR provides a basic implementation to fix #408.  I know the config commands are pretty self-explanatory, but this prevents the command line from displaying an error message.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,3 +64,6 @@ nosetests.xml
 doc/_build
 
 tests/shell_transcripts/*
+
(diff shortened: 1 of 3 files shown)


deargle:
Thanks again!

PR #413, opened 2020-04-06
🐛 fix #412: jquery path typo, localhost=>127.0.0.1, contact_email_on_error

### LINKED ISSUE #412: psiturk-example issues: email config error + jquery path missing '.' in '.min'
running `debug` form psitrk-example works up to launching from the ad. The pop-up window experiment throws js errors. The console complains:
> uniqueId is not defined
>
> <anonymous> http://localhost:22362/static/js/task.js:8
> var psiTurk = new PsiTurk(uniqueId, adServerLoc, mode);
defaults provide an unquoted address causing errors, adding quotes in the config are escaped before ending up in the js file -- same issue.

resolved by setting `config.ini` "contact_email_on_error = 0" 

> GET http://localhost:22362/static/lib/jquery-3.4.1min.js net::ERR_ABORTED 404 (NOT FOUND) 

using `jquery-3.4.1.min.js` or `jquery-min.js`  in `templates/exp.html` (adding missing dot or ignore viersion) seems to work.

and maybe related to #133, using `localhost` as the host causes issues for firefox (not a fully qualified domain name). using `127.0.0.1` instead seems to work.

WillForan:
I didn't follow `contact_email_on_error` to see if you should be a quoted address or is a boolean flag. This is just a quick fix so the example will run in a browser.

### DIFF
--- a/psiturk/default_configs/local_config_defaults.txt
+++ b/psiturk/default_configs/local_config_defaults.txt
@@ -7,7 +7,7 @@ us_only = true
 approve_requirement = 95
 number_hits_approved = 0
 require_master_workers = false
-contact_email_on_error = youremail@gmail.com
+contact_email_on_error = 0
 ad_group = Default psiTurk Stroop Example
 psiturk_keywords = stroop
 organization_name = New Great University
@@ -19,7 +19,7 @@ database_url = sqlite:///participants.db
 table_name = turkdemo
 
 [Server Parameters]
-host = localhost
+host = 127.0.0.1
 port = 22362
 cutoff_time = 30
 logfile = server.log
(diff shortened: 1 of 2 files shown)


deargle:
Oops thanks for jquery path fix. Artifact of not having a testing suite which can copy from the example files and run in a browser. Why was localhost => 127.0.0.1 necessary, and was the contact_email breaking something.

deargle:
To make it a boolean so that  javascript "If contact_email" doesn't display anything if it's not there?

deargle:
duh you opened an issue, I'll read that

deargle:
I added quote marks to the javascript in exp.html, it should fix the string parsing error now. Thanks for this PR.

WillForan:
oh duh ... quotes in exp.html seems obvious. thanks for the quick review!

Issue #414, opened 2020-04-07
config.txt: public experiments, private/heroku-specific Server Parameters?


WillForan:
I've an experiment tracked with git and pushed both to heroku and github. I think I want to keep config.txt in the repo (The instance running on heroku panic if it's not there?) but I don't want the secrets to be public.

I thought I could  use `custom.py` + env variables to overwrite parameters in `config.txt`. But that isn't loaded before the server is running(?). Is there already a way to set heroku specific server info?

```
# from 'host','port','login_username','login_pw','secret_key'
# to   'PSI_HOST' 'PSI_PORT' 'PSI_LOGIN_USERNAME' 'PSI_LOGIN_PW' 'PSI_SECRET_KEY'
for server_setting in config['Server Parameters']:
    envvar='PSI_' + server_setting.upper()
    envval=os.environ.get(envvar)
    if envval:
        config['Server Parameters'][server_setting] = envval
```

deargle:
You're running on master (dev branch), yeah? I have a desire to overhaul the setting of config vars -- to make more (all?) settable - overrideable via env vars.

The heroku instance will not panic, because the way that configparser works, is to first load all of the defaults, and then to override the defaults with settings from a project's config.txt. The only thing that heroku would balk about is that they set `$PORT` that web dynos _have_ to bind to in order to receive traffic. I don't know whether `$HOST` matters, whether it's 127.0.0.1 or 0.0.0.0. But heroku puts a proxy server in front of the gunicorn instance that is launched.

What are you using the login_username, login_pw, and secret_key for? I think that secret_key is basically unused IIRC but I could be way wrong, but otherwise it is used by flask for something or other for when it does secret stuff that we're not doing. login_username and login_pw I definitely want to make env-able, _in addition_ to which I want to perhaps only be setting a hash of the password.  As a safer default for users of psiturk who wouldn't know better than to not share their config.txt. But the config.txt often also contains database_url with secrets stored therein, which is another reason not to have a public config.txt. But that is overrideable via DATABASE_URL env var, because that's what heroku sets when the postgresql addon is... added on.

If we store the login_pw in the config.txt, passwords could be set by some command-line tools that accept plaintext input and then output a hashed secret, I think it's ansible-vault or something like that. But even then, yes it'd be better to override some things with env vars.

[...]

PR #415, opened 2020-04-07
look to env for config override

WillForan:
addresses #414

In poor form (sorry!) there are actually 3 things going on:
1. look to env for a lot of server settings
2. I think EphemeralContainerDBError wasn't imported. Might have caused a crash on heroku that was hard to debug (no logging?). Not sure if this will help.
3. Heroku doc notes for about DB and port


WillForan:
I think i muddied the git history. I'm not attached to history or authorship -- if any of it's useful, feel free to import it however works best for you

deargle:
Merged into PR #438 -- thanks!

PR #433, opened 2020-08-11
Heroku instructions warn about exposing database_url.

Tubeliar:
Commit instructions moved so people following the guide don't commit it by default. Added a note on ON_CLOUD and a googlable error you'd get if you forget to set it.


deargle:
Thanks! Yeah, I want to have more options for storing config params such as DATABASE_URL in places that aren't likely to get pushed to a public repo

PR #440, opened 2020-10-06
Fix bug in ExperimentServerController and improve cross-platform support

jacob-lee:
The method `kill_child_processes` had several bugs which apparently failed silently when running most of the time. However this method is only used to kill the child process of the shell process used to grep for the psiturk server processes in the method `is_server_running`.

I removed this `kill_child_processes` entirely, and altered the `is_server_running` method to avoid using a (platform dependent) set of operating system tools (grep, awk, ps). Instead it uses the cross-platform tools that shutil library already provides. In particular it uses `shutil.process_iter` to safely iterate through the processes in a way that avoids race conditions.

### DIFF
--- a/psiturk/experiment_server_controller.py
+++ b/psiturk/experiment_server_controller.py
@@ -140,47 +140,23 @@ class ExperimentServerController(object):
         else:
             self.server_running = False
 
-    def kill_child_processes(self, parent_pid, sig=signal.SIGTERM):
-        if os.uname()[0] is 'Linux':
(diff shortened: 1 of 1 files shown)


deargle:
Are you seeing the issue where `psiturk server restart` reports that the "experiment server may be already running..." after it tries to start it again, if it just killed it? I'm getting that both on master and also on this PR, so I don't think it's something you did.

jacob-lee:
I'm afraid I don't use the restart command. I usually rely on on/off

jacob-lee:
I don't know, but this sounds like a race condition sort of problem.

PR #442, opened 2020-10-07
Refactor server shutdown

jacob-lee:
- removes the http request and route to get ppid of server
- uses shutil to shutdown the process tree
- refactors the edge case for when ^C is entered in the psiturk-shell

### DIFF
--- a/psiturk/experiment.py
+++ b/psiturk/experiment.py
@@ -21,6 +21,7 @@ import re
 import json
 from jinja2 import TemplateNotFound
 from collections import Counter
+import signal
 
 # Setup flask
 from flask import Flask, render_template, render_template_string, request, \
@@ -58,6 +59,14 @@ logging.basicConfig(filename=LOG_FILE_PATH, format='%(asctime)s %(message)s',
 
 # Let's start
 # ===========
+
+# Unfortunately, this does not tell psiturk interactive to update server status
+def sigint_handler(signal, frame):
+    print('^C: shutting down server processes.')
+    sys.exit(0)
+
+
+signal.signal(signal.SIGINT, sigint_handler)
 
 app = Flask("Experiment_Server")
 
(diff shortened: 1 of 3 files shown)


deargle:
Hmm wait I'm confused, why have the flask server handle the ctrl+c exit? I sometimes run the flask debug server _without_ gunicorn, for debugging.

jacob-lee:
It just sets a signal handler in experiment.py that catches SIGINT and
calls sys exit. I'm not hooking into gunicorn afaik. I probably should have
separated this change from the rest of the commit, but having it report
that there was a ctrl-c whenever there was a shutdown seemed wrong. It was
doing that because I use SIGTERM to shutdown, which allows the on exit hook
to run. But I couldn't figure a way to check how the exit was triggered in
the exit hook.

deargle:
Oh! Thanks, that makes sense.

I'll play with this after I finish the doc overhaul I'm doing, I'm a laser rn.

deargle:
Can we switch this to use logger instead of print statements?

jacob-lee:
I like switching to a logger because we can use a streamHandler for the CLI and shell/script, but add other handlers like an HTTP handler in case we want to push logs to web app interface.

I accidentally pushed the last commit to github. Please disregard any errors that show up.

jacob-lee:
Any more issues that need to be addressed before pulling?

deargle:
I'm going to test that heroku etc still works for me before merging this

deargle:
Thanks for this! Looks much cleaner.

deargle:
I just pushed 99de42803261ea850cd116fba20ac9d291928083 moving the sigint handling out of experiment.py and into experiment_server.py. I remembered my use case -- to run the experiment.py using `flask run`, to get the benefits of live code reloading, no need to restart psiturk server. flask server was complaining about the signal binding being in that file, since experiment.py gets forked a lot by it, and signal binding should only be in the main thread.

Issue #443, opened 2020-10-09
code quality issues in psiturk_shell.py


jacob-lee:
I noted several issues with psiturk_shell.py, including, but not limited to the following:

- PsiturkNetworkShell docstring says extends PsiturkShell, but it doesn't extend PsiturkShell, ti extends, Cmd. PsiturkShell looks like it doesn't exist.
- PsiturkNetworkShell::do_quit (line 176) doc string says its over-riding do_quit for network clean up. But that isn't true
- PsiturkNetworkShell::do_quit (line 377) is defined a second time in the same class.
- PsiturkNetworkShell::hit_create (line 481). The regex has a warning about invalid escape sequence. Probably just needs to be prepended with r (to make regex)
- Most doc-strings should be converted to triple """ form
- self.poutput is not defined in `run`
- Unresolved reference i do_list_qualifications (line 779)
- PsiturkNetworkShell::do_worker: results may be used before assignment
- Potential issues with variables that can be references before assignment
- Multiple references to undefined variables

deargle:
Thanks for tidying this.

> PsiturkNetworkShell docstring says extends PsiturkShell, but it doesn't extend PsiturkShell, ti extends, Cmd. PsiturkShell looks like it doesn't exist.

That's correct. There used to be a "cabin mode" that PsiturkShell was used for, and then the "connected" mode that PsiturkNetworkShell was used for, but at some point I got rid of the cabin mode after I made it possible to execute commands without starting the interactive prompt. And yeah some functions got defined twice and my tests didn't catch them.

What are you using for code quality checks? Pycharm? I should set up something like that...

deargle:
And yeah, docstrings have been long neglected...

jacob-lee:
PyCharm. I used to do all my dev in Emacs, but I got tired with fiddling with it. PyCharm works very well (and handles the JavaScript stuff pretty well too). Also I discovered that they give out the pro version of all Jet Brain's IDEs free with an academic license.

PR #444, opened 2020-10-09
443-ps_cleanup

jacob-lee:
Intended to fix issue https://github.com/NYUCCL/psiTurk/issues/443

### DIFF
--- a/psiturk/psiturk_shell.py
+++ b/psiturk/psiturk_shell.py
@@ -9,25 +9,20 @@ try:
     from urllib.parse import quote_plus
 except ImportError:
     from urllib import quote_plus
-import urllib3
 import urllib3.contrib.pyopenssl
 from psiturk.utils import *
 from psiturk.models import Participant
 from psiturk import experiment_server_controller as control
-from psiturk.psiturk_statuses import *
 from psiturk.psiturk_config import PsiturkConfig
 from psiturk.version import version_number
 from psiturk.psiturk_org_services import PsiturkOrgServices
 from psiturk.amt_services_wrapper import MTurkServicesWrapper
 from .psiturk_exceptions import *
-import sqlalchemy as sa
 import webbrowser
 from docopt import docopt, DocoptExit
-from cmd2 import Cmd, EmptyStatement
+from cmd2 import Cmd
 from fuzzywuzzy import process
-import signal
 import certifi
-import datetime
 import random
 import string
 import os
@@ -55,6 +50,7 @@ try:
 except ImportError:
     import readline
 
+
 def docopt_cmd(func):
     """
     This decorator is used to simplify the try/except block and pass the result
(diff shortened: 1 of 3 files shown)


psiturk/psiturk_shell.py
```diff
@@ -262,19 +240,19 @@ def do_config(self, arg):
         else:
             self.help_config()
 
-    config_commands = ('print', 'reload', 'help')
-
+    # TODO: this method does not appear to be used; remove?
```
deargle, line 243: all of the `def complete_*` are used for tab completion by the shell

deargle:
This looks good. Anything we can do to get pep8 to not complain about the def complete_ functions, which are actually used? Or is pep8 not complaining about that?

jacob-lee:
Currently I only see the following:

PEP 8: E722 do not use bare 'except': This probably should be except KeyError
PEP 8: E722 do not use bare 'except': probably should catch ValueError and OverflowError
PEP 8: E265 block comment should start with '# ' -- this is commented out code
PEP 8: E265 block comment should start with '# ' -- this is commented out code
PEP 8 naming convention violation -- variable name is cmd_psiTurk (I missed fixing this one)

There were a number of declarations that didn't have PEP8's preferred line spaceing, and a lot of them used `'''comment'''` instead of `"""comment"""`. I fixed those.

I'll go ahead and fix those.

jacob-lee:
This is from code in hit_create. I'm thinking it is reasonable to put a cap on the reward that might be entered; it would probably be appreciated should someone accidentally enter a very large number.

We could choose an arbitrary threshold (e.g. very liberal $10,000
```
        if m is None:
            self.poutput('*** reward must have format [dollars].[cents]')
            return
        try:
            reward = float(reward)
            if not math.isfinite(reward) or reward < 0 or reward > 10000:
                raise ValueError
        except (ValueError, OverflowError):
            self.poutput('*** reward must be in format [dollars].[cents]')
            return
```

or we could fetch there amt balance:

reward <= (self.amt_services_wrapper.amt_balance()).data  # might need to wrap a float around that amt_balance call ...

which would involve a call to amt ...

What do you think?

deargle:
> I'm not sure if amt returns an error if you create a hit you don't have the balance for.

It does throw an error. The wrapper should already catch the error and return a result that has the exception bundled, and print the exception. I tried to move as much of the logic out of the shell and into the wrapper as I could, because I use the wrapper as an api of sorts for the web interface. Shell really is supposed to just do interactive shell-type things, and shunt rest of logic off to the wrapper. 

And the wrapper, in turn, doesn't do any actual amt calls -- that's in a different file. The wrapper is like a psiturk middleware abstraction of boto calls, interacts with psiturk db etc.

jacob-lee:
Okay, in that case I'll skip the check inside the try block.

jacob-lee:
Any further issues that need to be addressed before pulling?

deargle:
I'm apparently terrible with git, so I opened a new PR using commits from this one plus one small tweak, but I labeled this as hacktoberfest-accepted which I think means it will count, lmk if otherwise.

And, my pep8 is complaining about different things, which means we should commit some kind of pep8 config file to the repo, for standardization. Would also be herculean to do a sweep through the whole project, but later.

PR #448, opened 2020-10-21
Fixes example

afdreher:
This patch adds quotation marks around codeversion to ensure that it is properly interpreted as a string and updates the comments to remove duplication of 'a number indexing counterbalancing conditions.'


deargle:
thanks, but dejavu i thought we already fixed this somewhere

deargle:
Ah, no, that was another variable in that file that needed quotes. Thanks for this!

And beware, I just merged some _big_ changes into master that I haven't live-checked yet. But master is bleeding edge dev, so.

afdreher:
I'm mostly trying to rewrite parts of an earlier version to solve my problem (I need to do Instructions -> Experiment Part A -> Instructions -> Experiment Part B -> ...) but have been running into function scope issues.  They're my fault; I'm not a JS person.  Periodically, I like to check the master and example to see what's new and make sure I haven't broken anything *too* badly.

PR #449, opened 2020-10-22
code cleanup

jacob-lee:
This is provisional. Am having trouble getting all my pytests to run properly.

I started this revision, then a major refactor of the code was committed, and so I had to do a lengthy manual rebase.


psiturk/experiment_errors.py
```diff
@@ -165,9 +166,10 @@ def to_dict(self):
...
         return rv
 
+
 class InvalidUsageError(ExperimentApiError):
     
     def __init__(self, *args, **kwargs):
         if 'status_code' not in kwargs:
             kwargs['status_code'] = 400
-        super(InvalidUsage, self).__init__(*args, **kwargs)
+        super(InvalidUsageError, self).__init__(*args, **kwargs)
```
jacob-lee, line 175: or just `super().__init__(*args, **kwargs)`

psiturk/amt_services_wrapper.py
```diff
@@ -37,6 +37,7 @@
 from builtins import range
 from builtins import str
 from future import standard_library
```
jacob-lee, line 39: If we are only supporting > 3.5 then I think we only need `from __future__ import generator_stop` and `from __future__ import annotations`:


https://docs.python.org/3/library/__future__.html

feature | optional in | mandatory in | effect
nested_scopes | 2.1.0b1 | 2.2 | PEP 227: Statically Nested Scopes
generators | 2.2.0a1 | 2.3 | PEP 255: Simple Generators
division | 2.2.0a2 | 3.0 | PEP 238: Changing the Division Operator
absolute_import | 2.5.0a1 | 3.0 | PEP 328: Imports: Multi-Line and Absolute/Relative
with_statement | 2.5.0a1 | 2.6 | PEP 343: The “with” Statement
print_function | 2.6.0a2 | 3.0 | PEP 3105: Make print a function
unicode_literals | 2.6.0a2 | 3.0 | PEP 3112: Bytes literals in Python 3000
**generator_stop | 3.5.0b1 | 3.7 | PEP 479: StopIteration handling inside generators
annotations | 3.7.0b1 | 3.10 | PEP 563: Postponed evaluation of annotations**

deargle:
Sorry you had to do that big manual rebase after my merge. At same time, secretly I saw phew that I wasn't the one who had to do that rebase! 😁 But seriously, sorry for your pain

psiturk/amt_services_wrapper.py
deargle: sounds good. the `__future__` stuff was left in there when support for python3 was added. And yeah we only support minimum python 3.6

psiturk/experiment_errors.py
deargle: :+1: that was a holdover from needing to support python2 iirc

psiturk/amt_services.py
```diff
@@ -26,13 +23,13 @@ class AmtServicesResponse(object):
     """class AmtServicesResponse."""
 
     def __init__(self, status=None, success=None, operation='', message='',
-                 data={}, **kwargs):
```
deargle, line 29: good! looks like this was another mutable default arg bug that might have bit hard later

psiturk/dashboard/__init__.py
```diff
@@ -1,120 +1,133 @@
...
 
+
 def try_amt_services_wrapper(view):
     @wraps(view)
     def wrapped_view(**kwargs):
         try:
-            _ = services_manager.amt_services_wrapper # may throw error
-                                                      # if aws keys not set
+            _ = services_manager.amt_services_wrapper  # may throw error
+            # if aws keys not set
```
deargle, line 69: This second comment row is supposed to be a continuation of the comment immediately above it. shouldn't they be visually aligned?

psiturk/dashboard/__init__.py
```diff
@@ -1,120 +1,133 @@
...
-        if mode not in ['live','sandbox']:
+        if mode not in ['live', 'sandbox']:
             flash('unrecognized mode: {}'.format(mode), 'danger')
         else:
             try:
                 services_manager.mode = mode
                 session[SESSION_SERVICES_MANAGER_MODE_KEY] = mode
                 flash('mode successfully updated to {}'.format(mode), 'success')
-            except:
-                flash(str(result.exception), 'danger')
```
deargle, line 97: I have no idea what I was doing here with `result.exception`

psiturk/dashboard/__init__.py
jacob-lee: Ah, I think I didn't notice. The PEP8 issue is an in-line comment should have two spaces after the code. Its a little too rigid for how comments end up being used, or for multi-line comments such as this.

psiturk/models.py
```diff
@@ -1,40 +1,42 @@
...
 
 Base = declarative_base()
 Base.query = db_session.query_property()
 
-def object_as_dict(self, filter_these=[]):
-    return {c.key: getattr(self, c.key)
-        for c in inspect(self).mapper.column_attrs if c.key not in filter_these}
+
+def object_as_dict(self, filter_these: list = None):
+    if filter_these is None:
```
deargle, line 31: another mutable default problem! aaaah thank you!

psiturk/dashboard/__init__.py
jacob-lee: I noticed this this morning. I changed it to

```
except Exception as e:
    flash(str(e), 'danger')
```

psiturk/dashboard/__init__.py
jacob-lee: aligning them makes it complain about unexpected indentation. I'd say we just put in on one line. its super hard to get in-line comments to fit the 80 char limit.

jacob-lee:
Is the integration check stalled?

jacob-lee:
Looks like __import__.annotations isn't back-ported to Python 3.6. Am removing.

deargle:
Integration tests should rereun for each commit, even on pull requests, you can see details here https://travis-ci.org/github/NYUCCL/psiTurk/pull_requests

jacob-lee:
Thanks. it looks like it did run. Was in queue longer than I had expected.

jacob-lee:
@deargle Is psiturk-setup-example supposed to setup a working example task, or set up a template for a new task? The behavior seems to have changed, e.g. you have to rename the files, and if you just try to run out of the box, `psiturk debug` gives an exception because it cand find `ad_url_domain`.

deargle:
> @deargle Is psiturk-setup-example supposed to setup a working example task, or set up a template for a new task? The behavior seems to have changed. If you just try to run out of the box, `psiturk debug` gives an exception because it cannot find `ad_url_domain`.

Eh, this warrants discussion. I made a few pretty big changes here. Notably, yes, there are a few things in the config file that make it _not_ work out of the box -- a few config settings that have no defaults, for example, including the database url I think. It may not be the right approach, but I want to steer participants away from using sqlite by default just because it's there in the config file and they never change it. Same for the HIT title etc.

The dissonance is that people use psiturk-setup-example as a starting point for getting files for their _own_ experiment, whereas that command before set up an _explicit, complete_ experiment. I think that's kind of weird, am thinking about moving the full stroop demo to an experiment exchange since that's conceptually what it is, and then having `psiturk-setup-example` actually be more like a `psiturk-init` or `psiturk-new`, which copies over example configs that users have to complete.

As for `ad_url_domain` missing when participants run `psiturk debug`, sounds like a logic bug.

deargle:
Also! My latest big merge strips out the psiturk ad server, and any psiturk-org code that was in the codebase. `master` branch is truly bleeding-edge rn. Would like to be able to rely on unit tests to provide a guiding assurance that things are working, but yeah there's still coverage holes. I'll write one for ad_url_domain.

jacob-lee:
I think it makes sense to separate out:

psiturk-setup-example :: give a running example
psiturk-setup-template :: set up for new project

jacob-lee:
fyi, if you test `psiturk debug` on the master branch, you'll find that it errors out on 'config' named not defined. My pull request fixes that.

deargle:
> fyi, if you test `psiturk debug` on the master branch, you'll find that it errors out on 'config' named not defined. My pull request fixes that.

Okay, good to know. I'll write a unit test that shows that failling, and then merge in your thing, and step 3 -- profit.



> I think it makes sense to separate out:
> 
> psiturk-setup-example :: give a running example
> psiturk-setup-template :: set up for new project
> 
> (or psiturk-init or psiturk-new)

I need to mull over more. I'm thinking about how projects like jekyll (`jekyll new`) handle this. With the experiment exchange "out", `psiturk-setup-example` would basically just be doing a `git clone` of a stroop example somewhere. That's all the experiment exchange code was ever doing anyways. I like the idea of there existing a stroop example repo somewhere that can be browsed without having to run `psiturk-setup-example` -- that is not currently possible. Hmm.

jacob-lee:
ugh
psiturk.config.py

`return f"{ad_url_protocol}://{ad_url_domain}:{ad_url_port}:{ad_url_route}"`

is not right. I'll change to:

`return f"{ad_url_protocol}://{ad_url_domain}:{ad_url_port}/{ad_url_route}"`

deargle:
> ugh
> psiturk.config.py
> 
> `return f"{ad_url_protocol}://{ad_url_domain}:{ad_url_port}:{ad_url_route}"`
> 
> is not right. I'll change to:
> 
> `return f"{ad_url_protocol}://{ad_url_domain}:{ad_url_port}/{ad_url_route}"`

Ah, yeah.

jacob-lee:
PsiturkConfig.get_ad_url throws a PsiturkException when it can construct the ad url from the config file. psiturk_shell wasn't catching that exception because it was expecting it to return a falsy value in that case. I've changed code to catch the exception.

jacob-lee:
Okay. I'm giving my go-ahead on this pull.

THANKS.md
```diff
@@ -5,17 +5,17 @@ A number of people have contributed to psiTurk by reporting problems,
...
+Anna Coenen - https://github.com/annacoenen \
+Dave Eargle - https://github.com/deargle \
+Todd Gureckis - https://github.com/gureckis \
+Jessica Hamrick - https://github.com/jhamrick \
+Jacob Lee - https://github.com/jacob-lee  \
+Doug Markant - https://github.com/dmarkant \
+Jay Martin - https://github.com/jbmartin \
+John McDonnell - https://github.com/johnmcdonnell \
+Alex Rich - https://github.com/alexanderrich \
+Joseph Slote - https://github.com/JSlote
```
deargle, line 19: wat, this should be a bulleted list. i'll switch it.

deargle:
@jacob-lee want to submit a PR with your pep8 conf file? hacktoberfest ftw

deargle:
and thanks for this!

Issue #451, opened 2020-10-28
Failed test debug -p


jacob-lee:
I resolved my issues getting pytest to run on my dev machine. When I did run it, I found that it was failing one test in test_do_commands in `test_psiturk_shell with command` `['debug -p']`

```
>           assert not captured.err
E           AssertionError: assert not 'generating debug url using `ad_url` config var\n'
E            +  where 'generating debug url using `ad_url` config var\n' = CaptureResult(out='[\x01\x02psiTurk\x01\x02 server:\x01\x02off\x01\x02 mode:\x01\x02sdbx\x01\x02 #HITs:0]$ mode sandbo...bugKTTJBW&hitId=debugDPJK04&workerId=debugIIT1PF&mode=debug\n', err='generating debug url using `ad_url` config var\n').err
```

The issue is that calls to Cmd2.pfeedback in `psiturk_shell.py` were added recently:

```python
try:
    base_url = self.config.get_ad_url()
    self.pfeedback('generating debug url using `ad_url` config var')
except PsiturkException:
    self.pfeedback('`ad_url_*` config vars not set; using Server '
                              'Parameters host and port vars.')
    host = self.config.get('Server Parameters', 'host')
    port = self.config.get('Server Parameters', 'port')
    base_url = f"http://{host}:{port}/ad"
```

Because `get_shell` in `test_psiturk_shell` creates a shell where the `quiet` parameter is set to False.

So the options are: 
1. ignore it, 
2. change the test,
3. change Cmd2 to output non-essential feedback to stdout instead of stderr.
4. remove call to pfeedback

I think #3 is best option. I've implemented in a PR request that I'll post momentarily.

deargle:
Ah yeah, I wrote that test and then got distracted. But that test would have caught the earlier `debug -p` bug you found

deargle:
Closed by #452

PR #452, opened 2020-10-28
Issue 451: do_debug fix

jacob-lee:
Fixes issue https://github.com/NYUCCL/psiTurk/issues/451

- sets Cmd2 to output non-essential info (pfeedback) to standard output
- changes one call to pfeedback to perror
- cleans up minor PEP8 issue.

### DIFF
--- a/psiturk/psiturk_shell.py
+++ b/psiturk/psiturk_shell.py
@@ -926,10 +926,11 @@ class PsiturkNetworkShell(Cmd, object):
             port = self.config.get('Server Parameters', 'port')
             base_url = f"http://{host}:{port}/ad"
 
-        launch_url = base_url + "?assignmentId=debug" + \
-            str(self.random_id_generator()) + "&hitId=debug" + str(self.random_id_generator()) \
-                     + "&workerId=debug" + str(self.random_id_generator()
-                     + "&mode=debug")
+        launch_url = (f'{base_url}?'
+                      f'assignmentId=debug{self.random_id_generator()}'
+                      f'&hitId=debug{self.random_id_generator()}'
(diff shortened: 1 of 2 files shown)


deargle:
all hail f-string! Thanks for finding the ` self.feedback_to_output = True` option

psiturk/psiturk_shell.py
```diff
@@ -94,7 +94,7 @@ def amt_services_wrapper(self):
                     '- `psiturk server off`',
                     '- `psiturk debug -p`'])
                 message = '{}{}'.format(e.message, still_can_do)
-                self.pfeedback(message)
+                self.perror(message)
```
jacob-lee, line 97: Setting feedback_to_output = True caused another test to fail. Changing this to `perror` fixed. `pwarning` is also probably an option.

deargle: I'm okay with this being perror, conceptually

deargle: You never got `pytest` or `tox` working locally? `tox` is a good bet when pytest is misbehaving, because tox creates an isolated testing environment. I don't mind you pushing code to get tests to run, but local testing would be much faster development cycle for you.

jacob-lee: Pytest is working for me. 

It wasn't before. I eventually worked out that the development environment I was using had a non-canonical set of dependencies installed via pip; I forget exactly why. Creating new environments from scratch got it working fine. So I created one for py36, py37 and py38. 

Yeah, testing locally is faster for sure.

I'll try tox.

deargle:
Thanks for this!

PR #470, opened 2021-03-15
disable psiturk-heroku-config attach-psql

deargle:


deargle:
The currently-undocumented `psiturk-heroku-config` command no longer attaches a postgresql addon by default.
I'm calling this just a bugfix because (1) the docs assume that it _doesn't_ attach the psql database, and
(2) that command isn't even documented anyway. Will version-feature-bump when this feature is better
thought through.

Issue #473, opened 2021-03-15
tests problem: tests of psiturk_shell are basically no-op


deargle:
I just hit a problem with psiturk_shell.py's create_hit that surprised me because that function is ostensibly covered by a test case. Turns out that the test cases for psiturk_shell are basically useless because nearly nothing does `self.perror`, and the tests check for error output. Fixing the problem will require some extensive stubbing work for any shell call that makes an mturk call, which I'm not excited about after all the stubbing for amt_services_wrapper...

PR #474, opened 2021-03-15
fix bugs in psiturk_shell hit_create feedback

deargle:


deargle:
See #473 , psiturk_shell.py probably has other problems too, but basically we have no insight with current unit tests for that file

PR #475, opened 2021-03-16
switch order of quiet and aws_wrapper check

deargle:
so that wrapper doesn't get checked if quiet until a psiturk_shell 
function asks for it


deargle:
Again, see #473, we basically have no test coverage of psiturk_shell. more fixes.

Issue #482, opened 2021-03-24
Feature: Adding support for stream logging (only used when ON_CLOUD=1)


jmuchovej:
I was recently trying to debug different local application behavior vs Heroku-hosted behavior. (One of my custom routes was breaking on Heroku, but not locally.)

However, since `ON_CLOUD` will reset psiTurk's logging, this made it a nightmare toe debug.
https://github.com/NYUCCL/psiTurk/blob/d81d3532c4cf09970a70c901671d27ad3941df09/psiturk/experiment.py#L37-L41

My changes are detailed [here][changes]. Essentially, I replaced the enforced logging to a file to `logging.<type>Handler`s so that a `StreamHandler` could be added, [allowing for Heroku pull `psiTurk` logs, since they aggregate all `stdout` and `stderr` output][heroku-logging].

[heroku-logging]: https://devcenter.heroku.com/articles/logging#writing-to-your-log

[log_reset]: https://github.com/NYUCCL/psiTurk/blob/d81d3532c4cf09970a70c901671d27ad3941df09/psiturk/experiment.py#L38
[changes]: https://github.com/jmuchovej/psiTurk/blob/8d45b182f18db0eb44eb30a0df87a68d1638b2b0/psiturk/experiment.py#L36-L51

deargle:
Wait I'm confused, but I admit to being a noob with logging, and I know there's some head scratching stuff in psiturk rn with logging. At one point that logging was disabled so that the SECRET_KEY didn't get logged to the heroku logs, but on second thought there's no need to log that anyway. psiturk_server.py sets the logger to be stderr, and I thought there was code that bound the flask logger to gunicorn's when gunicorn is detected. How do your changes fit in with that? Do you see the part I'm talking about? Thanks for helping with this btw, again the current state is a mess.

jmuchovej:
Hmm... Let me roll back the experiment to a state where I should have the breakdown. Give me a few minutes. 😅 

I don't recall seeing anything from `gunicorn` nor `flask`, but will check back.

jmuchovej:
## Rollback
Okay, so, rolling back to a few commits before I implemented the change – I see only logs from the following:
- `heroku[router]`
- `heroku[web.1]`
- `app[web.1]`

Trying the erroneous path, I don't see Heroku's logs raising the error through their dashboard (`https://dashboard.heroku.com/apps/<app>/logs`).

## `psiturk_server.py`?
[...]

PR #483, opened 2021-03-24
Adds StreamHandler to support Heroku-like logging

jmuchovej:
**Note:** I haven't tested this outside of local `psiTurk` and Heroku.

See #482 for rationale. From my understanding, though, this ought to work with any `stdout`/`stderr` intercepting log platforms.

### DIFF
--- a/psiturk/experiment.py
+++ b/psiturk/experiment.py
@@ -33,30 +33,21 @@ from .user_utils import nocache
 CONFIG = PsiturkConfig()
 CONFIG.load_config()
 
+LOG_LEVELS = [logging.DEBUG, logging.INFO, logging.WARNING, logging.ERROR,
+logging.CRITICAL]
+LOG_LEVEL = LOG_LEVELS[CONFIG.getint('Server Parameters', 'loglevel')]
+
+logfile = CONFIG.get("Server Parameters", "errorlog")
+if logfile != '-':
+    file_path = os.path.join(os.getcwd(), logfile)
+    logging.basicConfig(filename=file_path, format='%(asctime)s %(message)s',
+                        level=LOG_LEVEL)
+
 # Let's start
 # ===========
 
 app = Flask("Experiment_Server")
-
-# experiment server logging
-if 'gunicorn' in os.environ.get('SERVER_SOFTWARE', ''):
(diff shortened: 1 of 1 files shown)


jmuchovej:
I'll make appropriate edits to the `CHANGELOG.md` and add `Closes #483` commits once approved. 😅

deargle:
I've been poking at this. You're right that the problem is when `logging.basicConfig(filename=None)` is passed -- reading the docs on basicConfig, it (and Flask) create a StreamHandler if no other handler is defined. But since the `filename` key is present -- even though it's `None` -- the StreamHandler doesn't seem to get created. I don't want to only check if ON_CLOUD is set though before not using a filehandler logger, I want to check if logfile or errorlog == `-` and then handle that separately -- because that is what gets passed to gunicorn to tell it to go to stderr/stdout.

I'm going to push some changes to your PR branch -- could you test them once I do?

jmuchovej:
I have an experiment currently running (I believe it'll be done in the next hour or so), but afterwards – I'd be happy to test. 🙂

jmuchovej:
Just an update, I should be free-and-clear of running experiments until next Tue/Wed, so I can test any changes you implement. 🙂

deargle:
@jmuchovej I just pushed a commit to your PR branch that defers all logging to gunicorn. I think this is cleaner, since heroku already seems to do things like override any gunicorn setting to use `-` for the access-log location via an env var. But logs are appearing twice for I can't figure out what reason though, but it is logging exceptions and also current_app.logger.whatever. Is this working for you?

deargle:
Actually I'm going to keep looking into this, I think your approach is better.

deargle:
Yeah, I prefer your approach. No duplicate logging. I circled back and made a small tweak to your PR. Where you explicitly set up a streamhandler if on cloud, I'm letting flask set up its default streamhandler if it isn't given any filename. I'm referencing `-` instead of ON_CLOUD.

Thanks for figuring all this out. Turns out that a lot of the advice I linked on the internet about integrating flask and gunicorn logging is wrong.

jmuchovej:
Awesome, glad to see this helped! (Sorry, I couldn't get to this earlier. If testing would be helpful to confirm behavior, lmk.) 🙂

PR #484, opened 2021-03-26
adds and option on startup to move to a different subfolder

gureckis:
currently the Heroku set up requires all the files to be in the top level folder.  however some times it would be convenient to have multiple experiments in a single github folder and to swap between them.  this adds an option (None by default) to let you configure this.  the tip to use it would be, on Heroku, to change the `heroku_app.py` file so that you pass a directory to `exp.launch()`.  For example, if your experiment was in ./exp1 then `exp.launch("./exp1")`.  the `config.txt` should be placed inside exp1 folder in this instance as well as static, templates and the custom.py file.

### DIFF
--- a/psiturk/experiment_server.py
+++ b/psiturk/experiment_server.py
@@ -32,10 +32,11 @@ class ExperimentServer(Application):
     Custom Gunicorn Server Application that serves up the Experiment application
     """
 
-    def __init__(self):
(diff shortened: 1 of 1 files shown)


gureckis:
I just verified this minor change works from this branch to run from a subfolder... so let 'er rip (and be ignored by most people)

PR #487, opened 2021-03-29
Fix scoping bug in download_datafiles

BlaiseRitchie:
The variable `outstring` defined in a `with` statement is used outside the scope of the `with` in `get_question_data` and `get_event_data`, resulting in question and event data both throwing "Error reading record" errors (the exception being "I/O operation on closed file") when running `download_datafiles` from the psiturk shell.

### DIFF
--- a/psiturk/models.py
+++ b/psiturk/models.py
@@ -125,8 +125,7 @@ class Participant(Base):
                             event["timestamp"]
                          )
                     )
-
-            return outstring.getvalue()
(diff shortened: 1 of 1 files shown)


deargle:
Thanks for this!

jacob-lee:
Yes, indeed, thank you. It looks like I introduced this bug last October.
I'm surprised that code inspection didn't flag.

Jacob

Issue #458, opened 2020-11-25
Feature proposal: support for manual Qualification Types


decodyng:
Not sure if this is the right protocol, but: for a project I'm working on, we needed support for attaching a manually-created Qualification Type to our HIT (just the attaching; assume you've done the actual creation of the Qualification elsewhere), and I've put together a (quite small line-change) branch in my fork that I believe is doing this successfully. Is this something the maintainers might be interested in as a PR?

deargle:
It's already done on master branch, actually.

decodyng:
So, I thought this, and closed the Issue, and then realized it isn't actually done in the way I'd need, with a numeric rather than binary Qualification (that is, the kind generated by a QualificationTest)

decodyng:
Based on the fact that the master branch only checks for `Exists` rather than any kind of `GreaterThanOrEqualTo` logic that you'd need to grade a qualification test of the kind we're making for our project

https://github.com/NYUCCL/psiTurk/blob/master/psiturk/amt_services.py#L321

jacob-lee:
This is actually supported in a limited way now in the dev version on
GitHub.

Jacob

deargle:
A workaround is to create a new binary qualification based on your
continuous qualification. I'm hesitant to add code complexity when that
flow is possible.

evankirkiles:
@deargle How would one go about doing that, creating a binary qualification based on an existing qualification? Could you point me to relevant documentation that says this is possible?

deargle:
Closed by https://github.com/NYUCCL/psiTurk/pull/493 (authored by @evankirkiles)

As for the question of creating a binary qual from a continuous one, I was thinking of a python boto3 script that does something like: 

1. create a binary qualification, starting with no workers assigned
2. fetch all workers with any setting for a given qualification
3. loop over those workers. if qualification value > than some value, grant them the binary qualification created in step 1

PR #493, opened 2021-04-11
Add custom MTurk qualification support

evankirkiles:
Currently, the PsiTurk custom qualification feature only allows for binary has-this-qualification-been-granted support. However, it's not hard at all to abstract the qualification logic to allow for supplying a JSON file of more advanced qualification requirements, which would allow users to provide their own MTurk-formatted qualification tests with performance thresholds directly to the HITs, like seen in [this blog post](https://katherinemwood.github.io/post/qualifications/). 

I've been using this updated version of PsiTurk in all of my recently created HITs and it works wonderfully alongside a script I built to upload custom qualification tests to MTurk and return their ID. I've seen this issue crop up before (https://github.com/NYUCCL/psiTurk/issues/458) and the code is so simple to fix it and saves so much work that I figure a PR would be beneficial for a lot of people. I can also add in format checking to make sure that the advanced_qualifications JSON file has the correct structure, but overall I definitely think this feature could only be a positive addition.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - user_utils.PsiTurkAuthorization should not allow empty username or password! (#492)
 
+### Added
+- Add custom MTurk qualification support (#493)
+
 ## [3.1.0]
 ### Added
 - ability to launch the experiment server from a subdirectory instead of just
--- a/doc/settings.rst
+++ b/doc/settings.rst
@@ -248,6 +248,22 @@ workers with that qualification already set will neither see your ad nor be able
 to accept your HIT. This is the recommended way of excluding participants who
 have performed other HITs for you from participating in your new HIT.
 
+.. _advanced_quals:
+
+advanced_quals_path
+~~~~~~~~~~~~~~~~~~~
+
+A path to a custom JSON qualifications file, where you can define your own
+MTurk qualification requirements, as seen in `advanced_quals.json.sample`__
+
+__ https://raw.githubusercontent.com/NYUCCL/psiTurk/master/psiturk/example/advanced_quals.json.sample
+
+:type: ``path``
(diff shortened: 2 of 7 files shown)


deargle:
Thanks for this! Format checking would be great -- what is the basic requirement anyway, a list of dicts?

evankirkiles:
Yeah, it uses an array of dicts. For example, this qualification requirement I'm using adds the qualification test "3V8AA37DQPTARZNHJHM1A5CEB6ROW1" to the HIT. It shows up in the qualification requirements as a test workers can take, and only allows workers who get 100% on it to continue on to the HIT.

```
[
    {
        "QualificationTypeId": "3V8AA37DQPTARZNHJHM1A5CEB6ROW1",
        "Comparator": "EqualTo",
        "IntegerValues": [100]
    }
]
```

Now that I think about it, an incorrectly formatted Qualification ID will raise an error regardless upon trying to submit the HIT to MTurk––it might be better just to allow all the checking to be done by AWS itself rather than try to do it beforehand and potentially incorrectly. In that case, all that really needs to be done here is adding an example of this feature's usage to the demo and docs.

deargle:
The error would be pretty obscure if a dict was passed instead of an array
of dicts -- can we just check that it's an array of dicts, and let boto
handle the rest? And yeah a short example for the docs would be good.

evankirkiles:
OK, added list-of-dicts type checking and a boilerplate example in the example/ folder.

deargle:
I'm pushing some changes to make it so that no advanced_quals path is set by default. I would push them directly to this PR, but that would be your master branch, so I pushed to this branch: https://github.com/NYUCCL/psiTurk/tree/advanced_quals can you test that and make sure it's still accepting your advanced quals? I don't want to go through setting up the unit tests to use dummy ones right now.

evankirkiles:
Okay, merged those changes––everything still good on my end with my own advanced quals.

deargle:
👍 thanks!

PR #498, opened 2021-05-05
Remove getting local HIT id's from Participants table

evankirkiles:
Speeds up querying local HIT id's vastly, especially on AWS accounts with many participants. For example, my dashboard loaded the local HITs in about 8 seconds before this commit, and in <1 after this commit. 

Doesn't seem like there would ever be a case in which Participants would have a HIT Id not in the HIT Id table anyways, unless you're talking about really old backwards compatibility to before the HIT table would be a thing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [unreleased]
+### Added
+- Add custom MTurk qualification support (#493)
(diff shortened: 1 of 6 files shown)


deargle:
Actually yeah, I remembered, needing the hitid search to work for people upgrading was the rationale for that check of the participants table. I'm nervous -- even though I added that functionality to master long ago, it didn't get into the pip release until like a month ago with the release of psiturk 3. 

I think we need some kind of db migrate script? Something like `psiturk migrate-db` that, in this case, would copy hitids over from the participants table into the hits table. I would call the command something ambitious like `psiturk upgrade`, but I'm not ready to think about the logic of updating users' config.txt files etc.

deargle:
Thanks @evankirkiles ! I'm not sure how to best document the need for legacy experiments to run `psiturk migrate db`. I guess I'll just wait for panicked support requests :-(

PR #502, opened 2021-05-12
warn if not do_scheduler; fix if do_scheduler false; more verbose api error handler; fix update campaign goal

deargle:
overrides flask-api's default `handle_error`, which would just return a 
500 error for any exception


deargle:
This PR bloated a bit. All changes:

- Scheduler is started "paused" if `do_scheduler` is `False`, so that tasks can still be
  created, modified, and destroyed (#502)
- Campaigns' "update goal" functionality fixed -- the associated task is updated with
  the new goal (#502)
- `/dashboard/campaigns` and `/dashboard/tasks` now flash a warning if `do_scheduler` is `False` (#502)

PR #504, opened 2021-05-13
Deferred datastring column for insane Participant query speedup

evankirkiles:
By deferring the datastring column, we can skip loading in of the trial data JSON cells (which are generally massive), and only query them into a model on their first explicit reference from a Participant object. See [here](https://docs.sqlalchemy.org/en/14/orm/loading_columns.html) for more.

This cut my many-row queries on the Participants table down from 40+ seconds each to literally under 1. yay


deargle:
🎉 🎉 🎉 Thanks @evankirkiles !

PR #507, opened 2021-05-19
Update dashboard.txt

flxsosa:
Updating dashboard.txt to instruct users to also set secret_key as a config variable if they want to enable the dashboard. Before this commit, secret_key is not mentioned as a needed variable, leading to a PsiturkException when you enable the dashboard without secret_key.

Please review whether you can reference the secret_key variable through :ref:`settings-secret-key`. I was not sure if this was available as a reference but was following the syntax of :ref:`settings-login-*`


deargle:
merged via 956ef9fa57cbd8ab3f18347180a64418dc3fe102

deargle:
Thanks! lol there wasn't a secret_key section in the settings.rst at all

PR #517, opened 2021-06-16
Fix SSTI vulnerability in ad and consent pages

BlaiseRitchie:
Fixed an issue where users could pass arbitrary Python code to be executed on the server to the mode HTTP arg

More information about this type of vulnerability: https://secure-cookie.io/attacks/ssti/

### DIFF
--- a/psiturk/experiment.py
+++ b/psiturk/experiment.py
@@ -383,9 +383,10 @@ def advertisement():
         # even have accepted the HIT.
         with open('templates/ad.html', 'r') as temp_file:
             ad_string = temp_file.read()
-        ad_string = insert_mode(ad_string, mode)
+        ad_string = insert_mode(ad_string)
         return render_template_string(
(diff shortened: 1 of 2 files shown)


BlaiseRitchie:
Here's an example of reading a server-side file from /etc using the default psiturk-setup-example project:
SSTI vulnerability example

deargle:
🎉 🎉 🎉 awesome, how did you find this?

BlaiseRitchie:
> 🎉 🎉 🎉 awesome, how did you find this?

A whitehat security researcher reported it to us

deargle:
I'm considering just stripping out all of the `insert_mode` calls etc. I think that was for maintaining compatibility between psiturk 1 and 2 -- I don't think any psiturk 1 template would work anymore anyway.

deargle:
But I'll merge this for now.

PR #524, opened 2021-06-30
Dashboard Facelift

evankirkiles:
This was mentioned in an issue so I figured I'd create a PR so more people are aware of a restyling of the dashboard I've been developing and using with my lab. It seeks to solve the overflowing layout problems and responsiveness that exist in the current dashboard, and bring web browser functionality up to par with the CLI's capabilities. It's been super useful so far, especially for our lab manager and members who are far more comfortable in a browser interface than in a shell.

Unlike the previous dashboard which used Vue.js, this one is pure vanilla JS, powered mostly by a data viewport module I built a couple years ago and revamped here in dbview.js and dbfilter.js. The central layout on each page is the database and filter viewport on which you can apply any column-based comparative filters you want, and then a more customized pane depending on the functionality of the page. Actions like approval and bonusing can either be performed individually or on the visible rows in the database viewport. Said actions are handled in modals, as seen in the screenshots attached.

It should be comfortably responsive, and has worked well on my phone, but I have yet to test on a higher resolution than my laptop.

Screen Shot 2021-06-30 at 7 51 03 PM
Screen Shot 2021-06-30 at 7 52 30 PM
Screen Shot 2021-06-30 at 7 54 40 PM
Screen Shot 2021-06-30 at 7 53 38 PM
Screen Shot 2021-06-30 at 7 55 14 PM
Screen Shot 2021-06-30 at 7 53 02 PM
Screen Shot 2021-06-30 at 7 53 12 PM

### DIFF
--- a/psiturk/dashboard/static/js/campaigns.js
+++ b/psiturk/dashboard/static/js/campaigns.js
@@ -105,10 +105,12 @@ let vm_campaigns = new Vue({
     },
     methods: {
         fetch_all_campaigns: function(){
+            this.loading = true;
             return fetch('/api/campaigns').then(response=>{
                 return response.json()
             }).then(json=>{
                 this.campaigns = json
+                this.loading = false;
             })
         },
         create_new_campaign: function(){
--- a/psiturk/dashboard/static/js/hits.js
+++ b/psiturk/dashboard/static/js/hits.js
@@ -59,15 +59,15 @@ Vue.component('hit-table-row', {
             <dl>
                 <dt>HitId</dt>
                 <dd>{{ hit_data.hitid }}</dd>
-                
+
                 <dt>Title</dt>
-                <dd>{{ hit_data.title }}</dd>            
+                <dd>{{ hit_data.title }}</dd>
             </dl>
         </td>
         <td>{{ hit_data.status }}</td>
         <td>
             <div class='dropdown'>
-                <button class='btn btn-sm btn-outline-dark dropdown-toggle' 
+                <button class='btn btn-sm btn-outline-dark dropdown-toggle'
                     type='button' data-toggle='dropdown'
                 >Action</button>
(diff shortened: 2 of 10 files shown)


deargle:
Few months late, but thanks for sharing this! I like the dashboard styling, and the housekeeping and new worker routes on api.py are cool too. Similar motivations to you, I'm partial to keeping the JS within a framework I already sort-of know (Vue.js). I'm going to work on merging in a lot of this though.

psiturk/dashboard/__init__.py
```diff
@@ -1,82 +1,134 @@
...
+            _ = services_manager.amt_services_wrapper  # may throw error if aws keys not set
+        except NoMturkConnectionError:
+            raise Exception((
+                'Dashboard requested, but no valid mturk credentials found. '
+                'Either disable the dashboard in config, or set valid mturk credentials -- '
+                'see https://psiturk.readthedocs.io/en/latest/amt_setup.html#aws-credentials . '
+                '\nRefusing to start.'
+                ))
+    login_manager.init_app(app)
+    with app.app_context():
```
deargle, line 122: this is cool

psiturk/api/__init__.py
```diff
@@ -277,20 +193,335 @@ def post(self):
...
+                hit_duration_hours=hit_duration_hours)
+            return _return, 201
+
+
+
+class ServicesManager(Resource):
+
+    # POST: Set the services manager mode
+    #  mode: live or sandbox, the desired mode
+    def post(self):
```
deargle, line 264: I'm uncomfortable with this -- I think the api shouldn't have any inherent concept of a session. The dashboard, on the other hand, can conceptually know about a session.

deargle: On the other hand, my argument above probably means that the api shouldn't know about any services_manager either.

PR #527, opened 2021-09-28
Make assignment of table name consistent with instructions in config

jacob-lee:
… config file

In `psiturk/default_configs/local_config_defaults.txt` it says that `table_name` is deprecated and to use `assignments_table_name` instead. It also says that if
both are set, `assignments_table_name` will be preferred over `table_name`.

This code should satisfy that contract.

### DIFF
--- a/psiturk/models.py
+++ b/psiturk/models.py
@@ -19,7 +19,8 @@ config.load_config()
 
 # The "table_name" config key is deprecated -- it will be replaced by
 # `assignments_table_name` in a future release.
-ASSIGNMENTS_TABLENAME = config.get('Database Parameters', 'table_name')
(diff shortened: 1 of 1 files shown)


deargle:
Good catch, thanks for this. Can you work the fix into this area?  https://github.com/NYUCCL/psiTurk/blob/363aa5d2a9aa1cfb7fb0f84a48bb934c7de6e2b6/psiturk/psiturk_config.py#L59-L99 

I wanted to keep all backwards-compat stuff to `psiturk_config.py`, and let the rest of the code be agnostic.

jacob-lee:
Any idea why the tests in test_amt.py would fail here? I don't see any obvious connection to the change i made.

You're right, psiturk_config.py is a better place for the fix. I'll close this pull request and do that.

PR #530, opened 2021-10-01
Extracting the API changes from #524

deargle:
First pass at extracting just the API changes from #524


deargle:
#524 had a lot of changes to the api route functionality that I don't want to merge. But I do want its WorkerList resource, and its download-datafiles helper routes. Might change them a bit, but this is a first pass at extracting the relevant changes from that PR.

PR #537, opened 2022-01-05
Adds bonus history functionality to wrapper, increases page size

evankirkiles:
The MTurk boto3 client has functionality for getting a list of all paid bonuses for a specified HIT or for a specified assignment id. I've implemented this over on my own dashboard and it's been really useful to our lab with keeping payments clear regarding who we've paid per hit, and how much––especially for experiments not run through PsiTurk. I assume it can't hurt to have the master amt_services_wrapper also have this functionality if someone wants to work with this in the future.

I also bumped up the page size so we don't need as many pagination requests to AWS.

### DIFF
--- a/psiturk/api/__init__.py
+++ b/psiturk/api/__init__.py
@@ -278,6 +278,15 @@ class TaskList(Resource):
         raise APIException(message='task name `{}` not recognized!'.format(data['name']))
 
 
+class BonusList(Resource):
+    def post(self):
+        data = request.json
+        hit_id = data['hit_id'] if 'hit_id' in data else None
+        assignment_ids = data['assignment_ids'] if 'assignment_ids' in data else None
+        bonuses = services_manager.amt_services_wrapper.get_bonuses(
+            hit_id=hit_id, assignment_ids=assignment_ids).data
+        return bonuses, 201
+
 api.add_resource(ServicesManager, '/services_manager', '/services_manager/')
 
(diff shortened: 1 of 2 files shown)


psiturk/api/__init__.py
```diff
@@ -278,6 +278,15 @@ def post(self):
         raise APIException(message='task name `{}` not recognized!'.format(data['name']))
 
 
+class BonusList(Resource):
+    def post(self):
```
deargle, line 282: Should this be GET, not POST? my understanding of rest apis is that post is reserved for creation of new records, and GET for retrieval.

evankirkiles: I thought about this for a bit, and it's very possible to make it into a GET request. However, we'd then be unable to query for multiple assignment IDs. 

Actually, now that I think about it, I guess that that fits better with the MTurk boto3 client functionality itself. We'd just have to remove the multiple assignment ID workflows altogether. Will do when i find the time.

deargle:
Did you mean to also commit modifications to amt_services_wrapper.py? My intention was that no api calls touch amt_services.py directly, but rather, that they all go through _wrapper.py, which also exclusively handles any psiturk database work (IIRC). The shell also exclusively uses the wrapper.

evankirkiles:
Oh, yes, I forgot. I'll add a wrapper function, move all the multi-assignment-id calls out of amt_services.py and into the wrapper function, and then restrict the API to GET requests that can take either a HIT Id or a single Assignment Id.

Issue #554, opened 2022-12-06
legacy dependency: GitPython


tamruta:
GitPython is not being used. It is a ghost now.

jacob-lee:
And there is an open severe security vulnerability associated with the dependency as well.

https://security.snyk.io/vuln/SNYK-PYTHON-GITPYTHON-3113858

Issue #547, opened 2022-03-07
postgres database not allowing other engines or dialects


gateslm:
This issue was started on the Google Groups. 

When using `postgres` database, you are unable to use other engines. 

Linked to me was the following [code snippet](https://github.com/NYUCCL/psiTurk/blob/822d0882ced0c5ccf698b108bdd1bdebef228048/psiturk/models.py#L67-L70) about using postgres. You can see it is searching exactly for `postgres://`, which will not allow for variations like `postgresql://`. 

Mentioned in the Google Group, potentially adding regex or checking for either. 

The documentation for postgres recommends `postgresql://`. 

By making it regex, it could allow for other engines and dialects. 

Relevant links:
- [postgres Documentation about Database URL](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING)
- [Relevant issue from sqlalchemy](https://github.com/sqlalchemy/sqlalchemy/issues/6083) 
   - [Refer to this specific comment for more details on their end](https://github.com/sqlalchemy/sqlalchemy/issues/6083#issuecomment-831654538)
- [About using postgres versus postgresql](https://stackoverflow.com/questions/37910066/heroku-sqlalchemy-database-does-not-exist)
- [sqlalchemy with Engine configurations with `postgres`](https://docs.sqlalchemy.org/en/14/core/engines.html#postgresql)

Let me know if you would like more information.

PR #557, opened 2022-12-16
Updates dependencies for Python 3.10 support

evankirkiles:
Had been running into many issues with packages on newer versions of Python (3.10+). This commit updates most, if not all, the packages to their latest versions to allow for 3.10. It also removes 3.6 from the `python-app.yaml`, as GitHub actions no longer supports 3.6 (which recently reached EOL) on ubuntu-latest and thus fails Github workflows. Most of the newer package versions similarly no longer support 3.6.

Additional small changes, mostly for fixing failing test cases:
 - In `psiturk/experiment_server_controller.py`: The `check_server_process_running` function was throwing exceptions in all `psiturk_shell` tests on my machine, as some processes returned in `psutil.process_iter` had already been terminated and thus `proc.as_dict` would just throw. I wrapped in a try/catch block to remedy, which matches desired functionality.
 - In `test_noaws.py`, the Github Action similarly fails to pass the last test of because pytest would disable socket use without explicitly specifying it using `@pytest.mark.enable_socket`.
 - In `experiment_server_controller.py`, I was only able to pass the `psiturk version` test by removing `sys.exit()` from the server when not connected to AWS MTurk. I'm actually quite iffy on this change, because the psiturk CLI shouldn't depend on the server being run to function. I'd love input from someone who understands the functionality a bit better than me.

Related PRs which this solves:
- Issue: https://github.com/NYUCCL/psiTurk/issues/554 (and the PR https://github.com/NYUCCL/psiTurk/pull/555)
- PR: https://github.com/NYUCCL/psiTurk/pull/543, as newer versions of Flask do not need `markupsafe==2.0.1`

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
 
     steps:
     - uses: actions/checkout@v2
--- a/psiturk/experiment_server_controller.py
+++ b/psiturk/experiment_server_controller.py
@@ -159,9 +159,12 @@ class ExperimentServerController(object):
     def check_server_process_running(self, process_hash):
         server_process_running = False
         for proc in psutil.process_iter():
-            if process_hash in str(proc.as_dict(['cmdline'])):
-                server_process_running = True
-                break
+            try:
+                if process_hash in str(proc.as_dict(['cmdline'])):
+                    server_process_running = True
+                    break
+            except:
+                pass
         return server_process_running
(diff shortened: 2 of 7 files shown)


evankirkiles:
Newer versions of SQLAlchemy requires `postgresql://` in the place of `postgres://`, hence commit https://github.com/NYUCCL/psiTurk/pull/557/commits/93b4697ae32420de15495c54d1016851fe05de4f. This further addresses the following issue, though we might want to log something as well here:
- https://github.com/NYUCCL/psiTurk/issues/547.

Unfortunately, Heroku enforces its own `DATABASE_URL` as `postgres://`, which is unchangeable, so can't simply instruct users to always use `postgresql://`.

jacob-lee:
Not sure why sys.exit() is used at all in psiturk_shell.py, tbh.

PR #563, opened 2024-07-03
use with_entities to pull only cond and counterbalance

yang-yangfeng:
surprised no one has run into this issue before - my task app was using ridiculous amounts of memory from querying participants just to assign cond/counterbalance

I guess others probably save smaller/simpler/more efficiently coded datastrings or collect data from fewer participants for each codeversion

using .with_entities to just grab the cond and counterbalance columns seems more efficient

### DIFF
--- a/psiturk/experiment.py
+++ b/psiturk/experiment.py
@@ -206,7 +206,7 @@ def get_random_condcount(mode):
                    Participant.status == CREDITED,
                    Participant.status == SUBMITTED,
                    Participant.status == BONUSED,
-                   Participant.beginhit > starttime)).all()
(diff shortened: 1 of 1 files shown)


jacob-lee:
Seems like a reasonable improvement (particularly because all that potentially heavy task data gets stored in that one `datastring` column).

Too bad there is no-one managing this project anymore to accept the pull-request.

gureckis:
> Seems like a reasonable improvement (particularly because all that potentially heavy task data gets stored in that one `datastring` column).
> 
> Too bad there is no-one managing this project anymore to accept the pull-request.

I added you to the repo in a way that should allow merging PRs

jacob-lee (approved):

